PR #1, opened 2018-01-11
Switch from glide to dep. Check in vendor/

ryboe:
* switch from glide to dep
  * dep is the new, modern vendoring tool from the Go community
  * Go projects are encouraged to move to `dep`
  * only tool that guarantees a flat dependency tree
    * nested `vendor/` directories cause insidious bugs
* check in `vendor/`
  * best practice
  * users can just `clone` and `go build`
  * past advice to avoid checking in `vendor/` no longer applies in the age of `dep`
  * downside: repo size goes from 304K to 880K

### dep crash course
```sh
# add a new dep. prune away unnecessary files. the prune command will be
# incorporated into `dep ensure` in a future release of dep, so you will soon
# be able to skip the second command.
dep ensure -v -add github.com/foo/bar
dep prune

# update existing deps
dep ensure -v -update
dep prune
```
`Gopkg.toml` is for _constraints_. It's the _input_ to the dependency solver. `Gopkg.lock` is the _output_ of the dependency solver. Don't touch `Gopkg.lock`.
[version constraint syntax here](https://github.com/golang/dep#semantic-versioning)

P.S. Since I can't open an issue...
```sh
go vet ./...
dnscrypt-proxy/dnscrypt-proxy/serversInfo.go:67: range var oldServer copies lock: main.ServerInfo
dnscrypt-proxy/dnscrypt-proxy/serversInfo.go:69: assignment copies lock value to serversInfo.inner[i]: main.ServerInfo
dnscrypt-proxy/dnscrypt-proxy/serversInfo.go:73: call of append copies lock value: main.ServerInfo
dnscrypt-proxy/dnscrypt-proxy/serversInfo.go:100: assignment copies lock value to serversInfo.inner[candidate]: main.ServerInfo
dnscrypt-proxy/dnscrypt-proxy/serversInfo.go:100: assignment copies lock value to serversInfo.inner[0]: main.ServerInfo
dnscrypt-proxy/dnscrypt-proxy/serversInfo.go:134: return copies lock value: main.ServerInfo
```


Gopkg.toml
```diff
@@ -0,0 +1,31 @@
+[[constraint]]
+  name = "github.com/BurntSushi/toml"
+  version = "~0.3.0"
```
ryboe, line 3: `~0.3.0` means `dep ensure -update` will only update to `0.3.x` versions.

.gitignore
```diff
@@ -3,9 +3,6 @@
 #*#
 *.swp
 
-# Vendor directory
-vendor
-
```
ryboe, line 8: Checking in `vendor/` increases the repo size, but usually not by much. `vendor/` is all text, which git compresses. The benefit is that people can `clone` and `go build`. You won't have to waste time debugging issues coming from contributors with incorrect package versions. This can happen easily because `go build` searches the entire `$GOPATH`. It's really easy to accidentally build with an incorrect package version. The `vendor/` system was intended to fix this. Unfortunately, it wasn't 100% usable until `dep` came along.

MOZGIII: You can use `dep prune` after `dep ensure` to reduce the size of `vendor/`

jedisct1:
This is fantastic, thank you so much! As you can tell, I’m very new to Go, so that kind of feedback is very valuable, thanks again!

The fact that a copy of all dependencies have to be shipped with every project feels a bit weird, but I will follow your recommendations.

PR #24, opened 2018-01-23
local resolvers sources

s-s:
Support for local resolvers list managed by external tools. cache_file and valid minisign_key settings are required (validations @ config.go) but not used. IMHO should be ok for beta.


jedisct1:
We can probably come up with something a little bit cleaner.

Maybe by making the URL optional. After all a local source is just an already existing cache file.

s-s:
Yes, indeed, have the same thoughts. However it may confuse the end-user & will require more patching. Waiting for stable codebase to make it.

PR #31, opened 2018-01-25
Fix systemd socket support

Zirkelite:
I have fixed some parts that caused this feature to not be functional. With these changes I can confirm dnscrypt-proxy with systemd sockets works on my machine.

I have changed `activation.Listeners(true)` to `activation.Listeners(false)` because otherwise the environment variables passed by systemd to the service are blanked and then the UDP sockets can not be detected afterwards. Also, listeners need to be checked to not be `nil` in order to not pass it to the listener functions, since doing so causes an error.


jedisct1:
Awesome!

Thanks a lot for your help!

PR #82, opened 2018-02-03
Update example-dnscrypt-proxy.toml

Sak94664:
Change all "http" to "https" for better security


jedisct1:
Hi,

This doesn't provide better security since this content is signed. Even if the file is downloaded manually, a signature is required and verified, and if the verification fails, the file is deleted without being loaded.

The reason the example uses HTTP is that some routers don't have a clock, and knowing the current time and date is required to load HTTPS resources.

PR #107, opened 2018-02-08
switched  to https version  when available

Bitti09:
all remaining  http urls gave me a connection refused or   pointed to a ssl cert for another domain.


jedisct1:
Thanks!

PR #111, opened 2018-02-09
Cache Plugin: return non-fixed TTL for cached entries

amitbl:
Using the cache plugin today I noticed an error with how TTLs are returned.

for example with this configuration:
```
cache = true
cache_min_ttl = 300
cache_max_ttl = 7200
```

and this command:
``dig github.com A``

github.com A record TTL is dynamic (between 5-100) so the expected TTL to be returned by dnscrypt-proxy is 300. However, the original value is returned instead making clients issuing additional queries (although they are returned from cache and not being sent out again to upstream)
This is more of a issue when dnscrypt-proxy is not being used as local resolver.

Also, dnsmasq TTL values are relative to the time it hit the cache for the first time:
``dig google.com A`` returns TTL of 130, if I will query dnsmasq for a second time after 30 seconds the result is 100

```
; <<>> DiG 9.11.2-P1 <<>> google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 52449
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1204
;; QUESTION SECTION:
;google.com.                    IN      A

;; ANSWER SECTION:
google.com.             130     IN      A       172.217.23.14
....
;; WHEN: Fri Feb 09 19:45:23 UTC 2018
```

```
; <<>> DiG 9.11.2-P1 <<>> google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 52449
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1204
;; QUESTION SECTION:
;google.com.                    IN      A

;; ANSWER SECTION:
google.com.             100     IN      A       172.217.23.14
....
;; WHEN: Fri Feb 09 19:45:53 UTC 2018
```

I'm not sure if this behaviour is actually defined in the RFC, but I think it's a good choice.
This patch is fixing both.

### DIFF
--- a/dnscrypt-proxy/dnsutils.go
+++ b/dnscrypt-proxy/dnsutils.go
@@ -85,22 +85,33 @@ func getMinTTL(msg *dns.Msg, minTTL uint32, maxTTL uint32, negCacheMinTTL uint32
 
 func setMaxTTL(msg *dns.Msg, ttl uint32, force bool) {
 	for _, rr := range msg.Answer {
-		if (ttl < rr.Header().Ttl) || force {
+		if ttl < rr.Header().Ttl {
 			rr.Header().Ttl = ttl
 		}
 	}
 	for _, rr := range msg.Ns {
-    if (ttl < rr.Header().Ttl) || force {
+	  if ttl < rr.Header().Ttl {
 			rr.Header().Ttl = ttl
 		}
 	}
 	for _, rr := range msg.Extra {
-    if (ttl < rr.Header().Ttl) || force {
+	  if ttl < rr.Header().Ttl {
 			rr.Header().Ttl = ttl
 		}
 	}
 }
 
 func updateTTL(msg *dns.Msg, expiration time.Time) {
-  setMaxTTL(msg, uint32(time.Until(expiration) / time.Second), true)
+
+	ttl := uint32(time.Until(expiration) / time.Second)
+
+	for _, rr := range msg.Answer {
+	    rr.Header().Ttl = ttl
+	}
+	for _, rr := range msg.Ns {
+	    rr.Header().Ttl = ttl
+	}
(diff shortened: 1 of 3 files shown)


jedisct1:
That totally makes sense, thanks!

The only minor thing I'm not a big fan of in this patch is the `force` parameter of `setMaxTTL`, because `setMaxTTL` doesn't describe what the function does any more.

Maybe introduce a private function called both by `update` and `setMax`? Given its size, it will be inlined by the compiler no matter what.

amitbl:
I don't see a way to do this without looping over the RR's both in the private function and setMaxTTL as well because of this condition: ``if (ttl < rr.Header().Ttl)``.
Maybe it's a good idea to leave setMaxTTL with no change and implement updateTTL like this?
```
func updateTTL(msg *dns.Msg, expiration time.Time) {

  ttl := uint32(time.Until(expiration) / time.Second)

  for _, rr := range msg.Answer {
      rr.Header().Ttl = ttl
  }
  for _, rr := range msg.Ns {
      rr.Header().Ttl = ttl
  }
  for _, rr := range msg.Extra {
      rr.Header().Ttl = ttl
  }
}
```

jedisct1:
Let's do the `updateTTL` function that way, then. Almost-duplicate code is fine as long as it's very short.

jedisct1:
Awesome!

Merged!

Thanks a lot for your contribution!

PR #120, opened 2018-02-11
feat: add a flag to setup the open URL timeout

bdossantos:
Hi!

Sometimes I randomly encounter a timeout when I generate blacklist. This commit add the
ability to increase the timeout delay (default to 10s).

Thank you,


jedisct1:
That's a good idea!

Merged!

Thanks for your contribution!

PR #149, opened 2018-02-19
Handle time-based blacklist in generate-domains-blacklist.py

u1735067:
* Read time-based blacklist from domains-time-restricted.txt by default ; domains in this file are considered whitelisted (from other blacklists)
* Modified `list_from_url` to `load_from_url` to avoid reading the `time_restricted` file twice (1 for output, 1 for whitelist)
* Create `domains-time-restricted.txt` with commented examples

### DIFF
--- a/utils/generate-domains-blacklists/domains-time-restricted.txt
+++ b/utils/generate-domains-blacklists/domains-time-restricted.txt
@@ -1,0 +1,4 @@
+## Time schedules in dnscrypt-proxy.toml
+# twitter.com @work
+# facebook.com @work
+# *.youtube.*  @time-to-sleep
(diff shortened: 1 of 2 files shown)


jedisct1:
I don't understand. How is it different from the existing whitelist?

u1735067:
"Handle time-based blacklist" that's the main point :)
"Whitelist time-based blacklist" is bonus to avoid to copy the list in the white list (unless you don't care a blacklist override your time-based blacklist; or the soft prioritize the time-based blacklist over the simple blacklist already?)

jedisct1:
Makes sense.

Merged, thanks!

PR #237, opened 2018-03-17
Fix 2.0.6 ipv6 attach port

bleeee:
My idea of the bug fix. Enough for the program to function but not enough for the defence of on purpose injection like "[127.0.0.1]" or "2001::1]".


jedisct1:
Awesome!

Your changes have been merged!

Thanks!

coolmans:
подскажите как собрать проект под винду 64 , скачал Visual Studio Code

PR #255, opened 2018-03-23
Update dnscrypt-proxy.service

steave-r0gers:
Safety improvements: work of service on behalf of the system user, restriction to file system.
--------------------
add to wiki:

## run as dnscrypt user
addgroup --system dnscrypt
adduser --system dnscrypt --shell /bin/false --no-create-home --disabled-password --ingroup dnscrypt --quiet

## set capability for listening privileged 53 port by user dnscrypt
setcap cap_net_bind_service=+ep /opt/dnscrypt-proxy/dnscrypt-proxy


jedisct1:
```sh
# adduser
-ash: adduser: not found
# addgroup
-ash: addgroup: not found
# setcap
-ash: setcap: not found
```

These instructions are great for people building packages for specific linux distributions, but not for the rest of us.

The systemd changes also look a bit too restrictive. Where are the cache files supposed to be stored? What changes to the example configuration file are missing to be compatible?
Even after reading the systemd file, I must confess that I have no idea :)

The default file should probably allow people to keep storing cache files, executable files, blacklists, log files, etc. wherever they want.

Most people don't have your linux skillset, and a default configuration file that only works under some preconditions will increase the number of "it doesn't work".

ghost:
> These instructions are great for people building packages for specific linux distributions, but not for the rest of us.

This instructions and changes are terrible and will fail on most distributions. No distro will want to install and run dnscrypt in `/opt`

> set capability for listening privileged 53 port by user dnscrypt
> 
> setcap cap_net_bind_service=+ep /opt/dnscrypt-proxy/dnscrypt-proxy

This doesn't make sense and is redundant when [AmbientCapabilities](https://github.com/jedisct1/dnscrypt-proxy/pull/255/files#diff-d2ed4509acf5f779cc0c185249fa98afR22) are set inside service. It can  also be security risk as anyone will have access to those privileges.

[addgroup and adduser](https://manpages.debian.org/unstable/adduser/adduser.8.en.html) are frontends to [useradd](https://manpages.debian.org/unstable/passwd/useradd.8.en.html) and [groupadd](https://manpages.debian.org/unstable/passwd/groupadd.8.en.html) and aren't available in all distros. Moreover [systemd-sysusers ](https://www.freedesktop.org/software/systemd/man/systemd-sysusers.html)with [sysusers.d](https://www.freedesktop.org/software/systemd/man/sysusers.d.html) is recommended way for adding new users/groups this days. It will be best to use [DynamicUser=yes](http://0pointer.net/blog/dynamic-users-with-systemd.html) option in systemd service but it needs at least systemd 232 and 235+ is recommended.

Here's [example](https://aur.archlinux.org/cgit/aur.git/plain/dnscrypt-proxy.service?h=dnscrypt-proxy-go) of nice and working service but it may not fit for all distros and users.

dvzrv (changes requested):
`DynamicUser=yes` instead of the huge amount of unneeded customization would make much more sense actually.

dvzrv:
@FedericoYundt  good example! Although why is there no `RuntimeDirectory`?

ghost:
@dvzrv I don't know. Maybe dnscrypt-proxy didn't need runtime directory access when someone who prepared this example tested it. I sent new PR which has it included but commented out among with other options.

PR #257, opened 2018-03-24
Let default systemd socket listen on both IPv4 and IPv6

RotoDhoulmagus:
Setting listen_addresses = [] in config will listen on systemd socket, but by previous systemd socket config it would only listen on IPv4 127.0.0.1:53 without IPv6. This change fixes it.


jedisct1:
Does that only listen to the local interface?

RotoDhoulmagus:
Oh right, that would be also listening to non-local interface, if that's not expected. Now this should be good.

PR #259, opened 2018-03-24
Add optional hardening to systemd service

ghost:
It's improved alternative to https://github.com/jedisct1/dnscrypt-proxy/pull/255 . `DynamicUser=yes` is commented out as I'm not sure if all targeted systems have needed systemd version.

It could be enabled by default in future. Users are supposed to [edit](https://github.com/jedisct1/dnscrypt-proxy/blob/master/systemd/dnscrypt-proxy.service#L17) this file themselves anyway so I hope this isn't a problem.


jedisct1:
Merged, thanks!

PR #260, opened 2018-03-24
dnscrypt-proxy.socket: update [Unit] dependencies to match those from service

ghost:
dnscrypt-proxy.socket can be activated alone without dnscrypt-proxy.service so make sure it has correct dependencies.


jedisct1:
Merged, thanks!

Issue #239, opened 2018-03-19
systemd: allow to run as dedicated user


dvzrv:
I might be mistaken, but at least with the previous dnscrypt-proxy (pre 2.0.0) it was possible to run the service as a dedicated user using systemd, not requiring root.

However, if one wants to bind ports below 1024, it is needed to set some additional things.
Also, it would be beneficial to use DynamicUser (so it will not be required to set up a new system user).

Here is an example of such a service

```
[Unit]
Description=DNSCrypt client proxy
Documentation=man:dnscrypt-proxy(8)
After=network.target
Before=nss-lookup.target
Wants=nss-lookup.target

[Service]
NonBlocking=true
DynamicUser=yes
RuntimeDirectory=dnscrypt-proxy
CapabilityBoundingSet=CAP_NET_BIND_SERVICE
AmbientCapabilities=CAP_NET_BIND_SERVICE
ExecStart=/usr/bin/dnscrypt-proxy

[Install]
WantedBy=multi-user.target
```
Also note: `Type=simple` is the default in the case of dnscrypt-proxy and can be omitted.

Additionally, it can be beneficial to have templated services (only start proxying to a specific server). I'm not sure, this is still useful with version > 2.0.0, but it was with version < 2.0.0.
Here is an example:

```
[Unit]
Description=DNSCrypt client proxy
Documentation=man:dnscrypt-proxy(8)
Requires=dnscrypt-proxy@.socket
After=network.target
Before=nss-lookup.target
Wants=nss-lookup.target

[Service]
NonBlocking=true
DynamicUser=yes
RuntimeDirectory=dnscrypt-proxy
CapabilityBoundingSet=CAP_NET_BIND_SERVICE
AmbientCapabilities=CAP_NET_BIND_SERVICE
ExecStart=/usr/bin/dnscrypt-proxy -R %i -S -Z "%p@%i"

[Install]
WantedBy=multi-user.target
```
It's unclear to me atm, if the flags are still those from before version 2.0.0.

As the templated service also requires a templated socket (but the socket can not be set dynamically), it's not necessarily needed to provide a file.
However, if one wanted to use a certain server on a specific port, it can be achieved in the following way:

* copying the following to `/etc/systemd/system/dnscrypt-proxy@myserver.socket`

```
[Unit]
Description=dnscrypt-proxy listening socket

[Socket]
ListenStream=127.0.0.1:53 #example port
ListenDatagram=127.0.0.1:53 #example port
[...]

PR #261, opened 2018-03-24
Adding DynamicUser to systemd service file, enhancing socket and service

dvzrv:
Adding nss-lookup.target to the socket Before and Wants directive. Adding current upstream wiki as documentation to service and socket file.

Adding DynamicUser=yes to the service file, alongside various hardening settings (Protect{ControlGroups,KernelModules}. Allowing the service to bind to ports below 1024 by setting CAP_NET_BIND_SERVICE. Adding {Cache,Logs,Runtime}Directory for dnscrypt-proxy. Removing (default) Type=simple. Adding a more default ExecStart location and usage of configuration.

This is in accordance with #239 and with a little more inspiration taken from the [downstream AUR package](https://aur.archlinux.org/packages/dnscrypt-proxy-go/).

Note: I have removed the explicit `Requires` and `Also` directives from the service, as enabling and starting a socket of the same name should be enough.

Using `DynamicUser=yes` prevents the need for packagers of this software to explicitely create a user for this service (which should not be needed).

### DIFF
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -26,7 +26,7 @@
 
 
 ## List of local addresses and ports to listen to. Can be IPv4 and/or IPv6.
-## To only use systemd activation sockets, use an empty set: []
+## Note: When using systemd socket activation, choose an empty set (i.e. [] ).
 
 listen_addresses = ['127.0.0.1:53', '[::1]:53']
 
--- a/systemd/dnscrypt-proxy.service
+++ b/systemd/dnscrypt-proxy.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=DNSCrypt client proxy
+Description=DNSCrypt-proxy client
 Documentation=https://github.com/jedisct1/dnscrypt-proxy/wiki
 After=network.target
 Before=nss-lookup.target
@@ -7,15 +7,20 @@ Wants=nss-lookup.target
 
 [Service]
 NonBlocking=true
+ExecStart=/usr/bin/dnscrypt-proxy --config /etc/dnscrypt-proxy/dnscrypt-proxy.toml
+ProtectHome=yes
+ProtectControlGroups=yes
+ProtectKernelModules=yes
+CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+
+# Run dnscrypt-proxy as unprivileged user with
+# temporary assigned UID/GID. See man:systemd.exec
+# for more info. Requires systemd 232+.
 DynamicUser=yes
 CacheDirectory=dnscrypt-proxy
 LogsDirectory=dnscrypt-proxy
 RuntimeDirectory=dnscrypt-proxy
-CapabilityBoundingSet=CAP_NET_BIND_SERVICE
-AmbientCapabilities=CAP_NET_BIND_SERVICE
-ExecStart=/usr/bin/dnscrypt-proxy --config /etc/dnscrypt-proxy/dnscrypt-proxy.toml
-ProtectControlGroups=yes
-ProtectKernelModules=yes
 
 [Install]
 WantedBy=multi-user.target
--- a/systemd/dnscrypt-proxy.service
+++ b/systemd/dnscrypt-proxy.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=DNSCrypt-proxy client
 Documentation=https://github.com/jedisct1/dnscrypt-proxy/wiki
+Requires=dnscrypt-proxy.socket
 After=network.target
 Before=nss-lookup.target
 Wants=nss-lookup.target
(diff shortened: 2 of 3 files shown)


ghost:
It seems to me that you didn't rebase to latest git master branch and in effect revert some previous changes: https://github.com/jedisct1/dnscrypt-proxy/commit/a95c7b729b04d1e1789925e0bb3c0c41563ce6b4 https://github.com/jedisct1/dnscrypt-proxy/commit/f2ab65aab77fadc669f3a7962f47d3b4572656aa https://github.com/jedisct1/dnscrypt-proxy/commit/bdc32cee90e099a49214c0b5315ea429959ea8c1

In effect you have removed [Protect-Home](https://github.com/jedisct1/dnscrypt-proxy/pull/261/files#diff-d2ed4509acf5f779cc0c185249fa98afL16), and [IPv6 support](https://github.com/jedisct1/dnscrypt-proxy/pull/261/files#diff-be1b1efaa687e767fe49b5fbe3b938afL8).

> Note: I have removed the explicit Requires and Also directives from the service, as enabling and starting a socket of the same name should be enough.

This will break when someone enables service instead of socket. In that case it's better to preload socket before starting unit than adding [capabilities](https://github.com/jedisct1/dnscrypt-proxy/pull/261/files#diff-d2ed4509acf5f779cc0c185249fa98afR15).

Also you can add second `Documentation` option instead of removing [existing one](https://github.com/jedisct1/dnscrypt-proxy/pull/261/files#diff-d2ed4509acf5f779cc0c185249fa98afL3).

I just checked and systemd [ignores unknown options](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) so we can safely enable `DynamicUser=yes` which will be no-op on older versions (but still we can keep [comment](https://github.com/jedisct1/dnscrypt-proxy/pull/261/files#diff-d2ed4509acf5f779cc0c185249fa98afL21) about that).

Considering above I recommend something like below:
```
[Unit]
Description=DNSCrypt client proxy
Documentation=man:dnscrypt-proxy(8)
Documentation=https://github.com/jedisct1/dnscrypt-proxy/wiki
After=network.target
Before=nss-lookup.target
Wants=nss-lookup.target

[Service]
NonBlocking=true
ExecStart=/usr/bin/dnscrypt-proxy --config /etc/dnscrypt-proxy/dnscrypt-proxy.toml
ProtectHome=true
ProtectControlGroups=yes
ProtectKernelModules=yes

# Run dnscrypt-proxy as unprivileged user with
# temporary assigned UID/GID. See man:systemd.exec
# for more info. Requires systemd 232+.
DynamicUser=yes
CacheDirectory=dnscrypt-proxy
LogsDirectory=dnscrypt-proxy
RuntimeDirectory=dnscrypt-proxy

[Install]
Also=dnscrypt-proxy.socket
WantedBy=multi-user.target
```
And leave dnscrypt-proxy.socket untouched.

dvzrv:
@FedericoYundt argh... I only did a fork and then created the feature branch. Bad timing I guess (definitely not intentional) :(

> This will break when someone enables service instead of socket. In that case it's better to preload socket before starting unit than adding capabilities.

I'm not exactly sure what you mean by that. What breaks, when only the service is enabled?
If `Also=dnscrypt-proxy.socket` is set, it explicitely also installs the socket file, if one wants to, or not (on enabling the service).
In any case, it is possible to use the service with and without the socket (but the socket can start the service, on access of the ports).
If one wants to use socket activation, the settings in the configuration file has to stay blank. If addresses/ports are set in the configuration file, they will (potentially) conflict with the settings in the socket (and one has to set both to achieve pariety).

If the socket file is enabled automatically, it would not make sense to have the `listen_addresses = []` exposed in the configuration file, as port configuration would solely depend on the .socket file.

I just tested and both running with and without the socket activation works.

> Also you can add second Documentation option instead of removing existing one.

This is correct. However, either I can't find it, or there just is no more man page in the repository (which seems logical, given the fact, that the go version is basically started from scratch).

I was under the assumption, that `DynamicUser=yes` sets `ProtectHome=true`, but [it actually only implies](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=) `ProtectHome=read-only`. The explicit `true` is probably a better choice here.

ghost:
I meant it would break if you didn't add `AmbientCapabilities` to service.  In my opinion socket activation is better choice than adding capabilities to service as it doesn't need additional privileges. If you enable socket activation you don't want those capabilities to be there. When someone want to change a port they can edit or override socket file. I don't see usecase when you want capabilities instead of socket.

> If the socket file is enabled automatically, it would not make sense to have the listen_addresses = [] exposed in the configuration file, as port configuration would solely depend on the .socket file.

This is exactly what is documented there:
```
## List of local addresses and ports to listen to. Can be IPv4 and/or IPv6.
## To only use systemd activation sockets, use an empty set: []

```

dvzrv:
Matter of preference I guess. I'd be fine with both!
Maybe the phrasing could be a little clearer in the configuration, though?

```
## When using systemd socket activation, choose an empty set: []
```

dvzrv:
@FedericoYundt alright, I've updated the pull request.
Should I also change the configuration description?

ghost:
Yes, you may clarify the wording. That being said, I'm not the one accepting PRs here. :smile:

dvzrv:
@FedericoYundt well, at least you seem to know systemd good enough and care about the topic ;-)
@jedisct1 I'll push a small change to the configuration description, but I think otherwise this will be good to go!

PR #274, opened 2018-03-28
Add support print ip addr of server

NgoHuy:
Because `-list -json` doesnot print ip addr, this commit to add it.
But the ip will contain port number.

Signed-off-by: Severus <huynhok.uit@gmail.com>

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -124,6 +124,7 @@ type ServerSummary struct {
 	Name        string `json:"name"`
 	Proto       string `json:"proto"`
 	IPv6        bool   `json:"ipv6"`
+	Addr        string `json:"addr"`
 	Ports       []int  `json:"ports"`
 	DNSSEC      bool   `json:"dnssec"`
 	NoLog       bool   `json:"nolog"`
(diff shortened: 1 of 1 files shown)


NgoHuy:
now the output looks like this
```
 {
  "name": "yandex",
  "proto": "DNSCrypt",
  "ipv6": false,
  "addr": "77.88.8.78:15353",
  "ports": [
   15353
  ],
  "dnssec": false,
  "nolog": false,
  "nofilter": true,
  "description": "Yandex public DNS server (anycast)"
 }
```

NgoHuy:
I still do not understand why DoH didnot have ipaddr

jedisct1:
Unfortunately that doesn't work for DoH servers. Maybe return an `addrs` array with all IPs and host names instead?

```
addrs: ['dns.cloudflare.com', '104.19.198.29', '104.19.199.29', '[2400:cb00:2048:1::6813:c71d]', '[2400:cb00:2048:1::6813:c61d]']
```

Resolving the actual IP addresses can be tricky, and irrelevant if they depend on the client IP. So we would probably only include IP addresses from the stamp (in addition to host names) if they exist.

jedisct1:
Because DoH use URLs, with host names. Stamps can include IP addresses, but they are optional.

jedisct1:
e09f0875c1c052fb57f00f16886a7f77d36a0f0b should do the trick

PR #330, opened 2018-04-08
minor

fantuz:
updated IETF draft link to version 5, updated the "official name" of the proposed protocol, which is indeed "DNS over HTTP" without S or 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # ![dnscrypt-proxy 2](https://raw.github.com/jedisct1/dnscrypt-proxy/master/logo.png?2)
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks!

publicarray:
I think that "DNS over HTTP" is an unfortunate name, It's missing the implied TLS and/or HTTP/2 part of the protocol. oh well. ¯\_(ツ)_/¯

jedisct1:
@publicarray It is a terrible name.

DNS-over-HTTP reads "insecure, and only accepts sequential requests".

Since the specification explicitly mentions that HTTP/2 must be used, along with the justifications, this is really dumb.

¯\_(ツ)_/¯

publicarray:
@jedisct1 I agree and I can't find where the the official name says it's "DNS over HTTP". All of the official material that I see with DoH includes the S.

PR #332, opened 2018-04-09
typo: http->https

Firminator:


publicarray:
@Firminator discussion/reason is in #330

Firminator:
OK thanks.

PR #341, opened 2018-04-11
Windows Command Scripts CleanUp

B00ze64:
RESTART now STARTS if the proxy is not already running (I see you changed it to Stop/Start; I went with a test where it will start if not already running and restart otherwise).
Single call to CD instead of calling in two places.
IF ERRORLEVEL instead of using the pseudo-variable.
No need to specify DNSCrypt-Proxy.exe path since current directory searched first.
ECHO OFF & Enable Extensions.
ECHO. instead of echo'ing double quotes (which actually prints the quotes).
New Stop script.


jedisct1:
Merged, thank you!

PR #350, opened 2018-04-12
Tiny typo in error message

B00ze64:
Correct tiny typo in "Unable [to] use source" error message.


B00ze64:
PS: Not quite sure how to make it forget all previous commits, so the commit history shows changes that have already been pulled. The only change is in Config.go...

jedisct1:
Thank you!

PR #361, opened 2018-04-14
Add dnscrypt-stamper CLI tool

gdm85:
This PR does the following:

* split `ServerStamp` in its own package
* add a couple unit tests for `ServerStamp`
* add a new CLI tool `dnscrypt-stamper` that helps at creating/viewing [DNS stamps](https://github.com/jedisct1/dnscrypt-proxy/wiki/stamps).

# Rationale

I found that the stamps are unfriendly (they obfuscate all of their content) and the generator at https://dnscrypt.info/stamps/ could use some more documentation/help text; by writing these tests and the CLI tool I could understand better how they work and thought to share the result upstream in order to help others too. As an example, the tool can be used to automate stamps creation.

Another important feature of having `ServerStamp` separated in a package is that other Go software can import it and use it straight ahead (not possible when the package is called `main` as it is now, that would invite copy/pasting instead).

Furthermore, people might just read the unit test to understand how a stamp is generated.

**NOTE:** The expected stamps in the unit test have been generated with https://dnscrypt.info/stamps/ and they do match nicely

# Features of dnscrypt-stamper

dnscrypt-stamper is purely a CLI tool and it allows to:
* generate a dnscrypt or DoH stamp
* visualize multiple stamps

The visualization obviously can serve also as validation.

# Example of output
dnscrypt-stamper without any argument or with `--help` or `-h` will print:
```
Usage of dnscrypt-stamper:
  --dnscrypt
    	create a dnscrypt stamp (default true)
  --dnssec
    	enforce DNSSEC (default true)
  --doh
    	create a DNS-over-HTTPS stamp
  --hashes string
    	SHA256 hashes for the DNS-over-TLS server certificate, in hexadecimal format (12:34:aa:bb:...), comma separated; colons are optional
  --host string
    	host name for DNS-over-TLS
  --ip string
    	IP address
  --no-filter
    	enforce no filter (default false)
  --no-log
    	enforce no logs (default false)
  --path string
    	path for DNS-over-TLS queries (default "/dns-query")
  --port uint
    	port for dnscrypt or DNS-over-TLS; if non-standard will be appended to host/provider name
  --provider-name string
    	provider name for the dnscrypt server; same as --host
  --provider-public-key string
    	provider public key fingerprint hash (SHA256) for the dnscrypt server, in hexadecimal format (12:34:aa:bb:...)

Specify one or more DNS stamps on command-line to visualize them:

dnscrypt-stamper sdns://AQcAAAAAAAAACTEyNy4wLjAuMSDDhGvyS56TymQnTA7GfB7MXgJP_KzS10AZNQ6B_lRq5BkyLmRuc2NyeXB0LWNlcnQubG9jYWxob3N0
sdns://AQcAAAAAAAAAAAAA
```

# TODO

I see some aspects of this PR that need improvement, I am submitting it anyway to get @jedisct1's feedback:

* documentation updates missing
* dnscrypt-stamper is not using `dlog` (can be a pro/con, debatable)
* did not update the vendor'd dependency via Gopkg, but commits are separated so it should be easy to do that

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 	"github.com/jedisct1/dlog"
+	"github.com/jedisct1/dnscrypt-proxy/stamps"
 )
 
 type Config struct {
@@ -368,11 +369,11 @@ func ConfigLoad(proxy *Proxy, svcFlag *string) error {
 func (config *Config) printRegisteredServers(proxy *Proxy, jsonOutput bool) {
 	var summary []ServerSummary
 	for _, registeredServer := range proxy.registeredServers {
-		addrStr, port := registeredServer.stamp.serverAddrStr, DefaultPort
+		addrStr, port := registeredServer.stamp.ServerAddrStr, stamps.DefaultPort
 		port = ExtractPort(addrStr, port)
 		addrs := make([]string, 0)
-		if registeredServer.stamp.proto == StampProtoTypeDoH && len(registeredServer.stamp.providerName) > 0 {
-			providerName := registeredServer.stamp.providerName
+		if registeredServer.stamp.Proto == stamps.StampProtoTypeDoH && len(registeredServer.stamp.ProviderName) > 0 {
+			providerName := registeredServer.stamp.ProviderName
 			var host string
 			host, port = ExtractHostAndPort(providerName, port)
 			addrs = append(addrs, host)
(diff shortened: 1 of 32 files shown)


jedisct1:
This is great!

Having something that parses and produces stamps as a standalone thing is a very good idea. It probably shouldn't be part of `dnscrypt-proxy` at all, since stamps can be defined for protocols that `dnscrypt-proxy` doesn't support.

`dnscrypt-stamper` should probably be a standalone project as well (and maybe just be called `dns-stamper`).

gdm85:
Glad it might be of use :)

The separation of the `stamps` package is made up of the commits 3da66073f9cdcd00d854ad28feeaffaf140b152b b44c03c30ad9e128f90f0c9b0809c8bf67ce8aed and 1ae53e8af0e16bcc65bd5fbe7629af9503930aec

The remaining three commits d5d84be94e76454100907b7ed52991e20fc12d2e 9188da603932387a15719cfac1d968a6bf6ccdc4 35ecf1707ca6f77850f4b0902efc29d48ae878dd constitute the CLI tool.

How would you like to proceed? Want to do the honors and cherry-pick them as appropriate?

jedisct1:
The stamp package was moved here: https://github.com/jedisct1/go-dnsstamps

gdm85:
Great! I will submit a PR there for the unit tests and create a repository with the CLI tool, that you can probably fork if you feel like maintaining it.

PR #387, opened 2018-04-18
chore: use malwaredomainlist hosts.txt file instead of CSV

bdossantos:
Hello,

The CSV download URL of "Malwaredomainlist" (http://www.malwaredomainlist.com/mdlcsv.php?inactive=off) is often in timeout when we try to generate the dnscrypt-proxy blacklist.

"Malwaredomainlist" also provide an URL to download a pre-generated hosts.txt file which seem to be more resilient and fast.

See also:

* https://www.malwaredomainlist.com/forums/index.php?topic=3270.0
* https://www.malwaredomainlist.com/hostslist/hosts.txt
* http://www.malwaredomainlist.com/mdlcsv.php?inactive=off

Thank you.

### DIFF
--- a/utils/generate-domains-blacklists/domains-blacklist.conf
+++ b/utils/generate-domains-blacklists/domains-blacklist.conf
@@ -40,7 +40,7 @@ https://mirror1.malwaredomains.com/files/justdomains
 https://ransomwaretracker.abuse.ch/downloads/RW_DOMBL.txt
 
 # Malware Domain List
-http://www.malwaredomainlist.com/mdlcsv.php?inactive=off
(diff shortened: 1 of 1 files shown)


jedisct1:
If I remember properly, the problem with the `hosts.txt` file is that it's very inefficient: it's a list of host names, not domain names.

So, it has a lot of overlaps, only blocks a limited set of names, thus is useless for anything using subdomains.

bdossantos:
Oh sorry, I didn't see this nuance between both files. In this case we can close this PR :-)

I'll try to increase the timeout and investigate why this url fetch fail most of time when I try to generate the blacklist on Travis and sometimes on my laptop.

PR #389, opened 2018-04-18
Wait for network connectivity under systemd

### LINKED ISSUE #295: Linux: how to start the proxy after the network is up?
Hey again,
I noticed that DNSCrypt starts scanning/testing/rating the resolvers before my laptop is actually connected to the network. This means that the first 3-5 tested servers will always be "timed out".  

This only happens when I wake up my laptop from suspend/sleep mode.

Archlinux, connecting to WiFi 5GHz.
Which .logs can I provide for helping?

mr-bo-jangles:
Currently the proxy is started with systemd before the network is ready to make DNS requests. This can cause issue with the proxy not resolving any requests well after the network coming up. This alteration addresses that in a rudimentary way.

Fixes #295


jedisct1:
Thanks!

PR #405, opened 2018-04-25
Replace unsigned branch instruction with signed branch instruction

aead:


jedisct1:
Did you see the length go negative? How can that happen?

aead:
In Go the length of a slice is a `int` not a `uint`. I've no idea why `JB` does not cause a jump - as you said it would be **very** weird if the length would turn negative. However the code fails because it read/writes out of bounds. (`XOR` macro). The only reason (I can see) is that it does not jump if the length goes under `64`. Also the Go assembler may generate machine-specific instructions from Plan9 asm - maybe there is a bug in the assembler.

PR #411, opened 2018-04-27
systemd: Add documentation to dnscrypt-proxy.socket

ghost:


jedisct1:
Is there a way to apply these options only to the TCP socket, or is systemd utterly broken to the point that they even didn't think about this?

ghost:
Basing on [systemd.socket](https://www.freedesktop.org/software/systemd/man/systemd.socket.html) manual I can think about 3 possible solutions:

1. **Use 2 different sockets**: `dnscrypt-proxy-tcp.socket`, `dnscrypt-proxy-udp.socket`. We can adjust service to enable them both with single `systemctl enable dnscrypt-proxy.service` command.
2. **Remove tcp only options**. I think they aren't required for dnscrypt-proxy to work, users may enable them themselves when needed.
3. **Keep status quo**. There will be errors in logs but things will still work.

I can adjust this PR if you tell me which solution is preferred or if you have some other idea.

Issue #398, opened 2018-04-22
intalled the latest version of dnscrypt-proxy V2, some issues


jpgpi250:
- Whenever I reboot my system (raspberry pi), the following error appears in the dnscrypt-proxy log:
**[ERROR] dial udp 176.56.237.171:443: connect: network is unreachable**

-I'm running dnscrypt-proxy on raspberry pi on address 127.10.10.1, port 5551. In the dnsmasq log, the following error is logged:
**dnsmasq[501]: nameserver 127.10.10.1 refused to do a recursive query**

when the command ./dnscrypt-proxy restart is executed, the following error is displayed:
**[FATAL] listen udp 127.10.10.1:5551: bind: address already in use**

mibere:
Reproducible, especially after a complete system reboot.

dnscrypt-proxy 2.0.10, Raspbian/Debian 9.4

dnscrypt-proxy.log:

> [2018-04-22 11:27:03] [NOTICE] Source [public-resolvers.md] loaded
> [2018-04-22 11:27:03] [NOTICE] dnscrypt-proxy 2.0.10
> [2018-04-22 11:27:03] [NOTICE] Now listening to 127.0.0.1:7053 [UDP]
> [2018-04-22 11:27:03] [NOTICE] Now listening to 127.0.0.1:7053 [TCP]
> [2018-04-22 11:27:03] [NOTICE] Now listening to [::1]:7053 [UDP]
> [2018-04-22 11:27:03] [NOTICE] Now listening to [::1]:7053 [TCP]
> [2018-04-22 11:27:03] [NOTICE] [dnscrypt.me] TIMEOUT
> [2018-04-22 11:27:03] [ERROR] dial udp 94.199.213.171:443: connect: network is unreachable
> [2018-04-22 11:27:03] [NOTICE] dnscrypt-proxy is waiting for at least one server to be reachable
> [2018-04-22 11:27:13] [NOTICE] [dnscrypt.me] OK (crypto v2) - rtt: 32ms
>

jedisct1:
Looks like the server was started before the network was available.

The root cause doesn't seem to be dnscrypt-proxy, but the way it is started.

jpgpi250:
I followed the installation instructions for pihole on the [wiki](https://github.com/pi-hole/pi-hole/wiki/DNSCrypt-2.0), they state:
./dnscrypt-proxy -service install: install dnscrypt-proxy service
[...]

Issue #288, opened 2018-03-30
TCP_NODELAY required?


geekmiki:
I am running dnscrypt on Archlinux and dnscrypt-proxy.socket reports the following:
Mar 30 01:50:34 RPi-SSHGateway systemd[1]: dnscrypt-proxy.socket: TCP_NODELAY failed: Protocol not available

Is there anything to do outside of dnscrypt-proxy to enable tcp_nodelay system wide?

Thank you in advance.

Utini2000:
The dnscrypt-proxy.service (or was it the .socket?) has a "no-delay" parameter inside. Maybe this is it?

geekmiki:
it is enabled in  dnscrypt-proxy.socket:
NoDelay=true

jedisct1:
It's very weird that the option cannot be set. `TCP_NODELAY` has been implemented on all Unix systems for 20 years. Are the UDP and TCP sockets in the correct order? Maybe they have been swapped?

If you don't absolutely need to use systemd sockets, don't use them. Delete the `.socket` file and just configure the IP addresses you want to use in the `dnscrypt-proxy.toml` file.

geekmiki:
TCP and UDP socket were not swapped.

I am trying without .socket now.

What is the format of the IP address in dnscrypt-proxy.toml? 127.0.2.1:5754 gives me an error:
Near line 31 (last key parsed 'listen_addresses'): Invalid float value: "127.0.2.1"

geekmiki:
Running without socket works fine. Thanks for your help.

Utini2000:
@jedisct1 but enabling the service will automatically also enable the socket?

> sudo systemctl enable dnscrypt-proxy.service
Created symlink /etc/systemd/system/multi-user.target.wants/dnscrypt-proxy.service -> /usr/lib/systemd/system/dnscrypt-proxy.service.
Created symlink /etc/systemd/system/sockets.target.wants/dnscrypt-proxy.socket -> /usr/lib/systemd/system/dnscrypt-proxy.socket.

geekmiki:
@Utini2000 You have to edit the dnscrypt-proxy.service unit and remove the .socket dependency (at least that worked for me)

dvzrv:
First of all:
`NoDelay=True` is a [non-default setting](https://www.freedesktop.org/software/systemd/man/systemd.socket.html#NoDelay=) for a socket unit.

[...]

PR #416, opened 2018-04-29
Re-enable systemd socket activation

ghost:
Without socket activation and DynamicUser=true, dnscrypt-proxy daemon can't bind to port below 1024 which will results in unit failures. See discussion in https://github.com/jedisct1/dnscrypt-proxy/pull/261 .

The current design inspired by Archlinux users and mantainers (it's the only distro that officially supports dnscrypt-proxy) really works. The main benefit is allowing dnscrypt-proxy daemon to run as completely unprivileged process. Socket activation doesn't affect speed and latency. The only thing users should know is to change listening ports in socket file instead of config file. I looked in past issues and reported problems regarding sockets were about TCP_NODELAY failures which were false positive and irrelevant.

I improved documentation in systemd files and wiki as much as I could to make crystal clear what expected behavior is. I believe it will help anyone who don't refuse to read the docs. The situation will improve when other distros adopt dnscrypt-proxy and users won't have to deal with it on their own.

### DIFF
--- a/systemd/dnscrypt-proxy.service
+++ b/systemd/dnscrypt-proxy.service
@@ -2,11 +2,11 @@
 Description=DNSCrypt-proxy client
 Documentation=https://github.com/jedisct1/dnscrypt-proxy/wiki
 
-## Use systemd sockets - Do not enable unless you are very familiar with the systemd socket activation mechanism.
-## Always try without systemd sockets before reporting any issues related to speed, latency or reliability.
-## If you enable this, the `listen_addresses` list in the main configuration file can be empty. Or, at least,
-## the addresses it contains should not overlap with the systemd socket addresses.
-# Requires=dnscrypt-proxy.socket
+## Use systemd socket activation. If enabled, the `listen_addresses`
+## list in the main configuration file should be empty. If you want to change
+## listenning address modify dnscrypt-proxy.socket instead.
+Requires=dnscrypt-proxy.socket
+After=dnscrypt-proxy.socket
 
 ## Start service after the network is online.
 ## Requires a wait service such as NetworkManager or systemd-networkd.
@@ -39,9 +39,9 @@ ProtectControlGroups=yes
 ## Deny explicit module loading.
 ProtectKernelModules=yes
 
-## Run dnscrypt-proxy as unprivileged user with
-## temporary assigned UID/GID. See man:systemd.exec
-## for more info. Requires systemd 232+.
+## Run dnscrypt-proxy as unprivileged user with temporary assigned UID/GID.
+## See man:systemd.exec for more info. Requires systemd 232+. If you don't
+## use systemd socket activation then you have to disable that option.
 DynamicUser=yes
 CacheDirectory=dnscrypt-proxy
 LogsDirectory=dnscrypt-proxy


jedisct1:
Ok, I give up with systemd. Its configuration files have nothing to do in this repository.

ghost:
This is really unfortunate. A lot of work was put into systemd integration which could benefit normal linux users and serve as a guidance for linux distro maintainers. After recent changes it was extensively documented as well. I was ready to help with any concern you may have with current implementation but you didn't provide me anything so I have no idea what the problem was.

Still having a `contrib` folder in this repo with community provided but not officially endorsed enchantments like systemd integration will be very valuable.

jedisct1:
If the only reason to use these activation sockets is to run the server as non-root on Linux, implementing privileges drop in the server itself would be way more valuable.

ghost:
No, running server without privileges from the start has more security benefits than starting it as root and then drop privileges. It doesn't need touching the code and trusting it does the right things.

Starting it as root also rules out many sandboxing features which can be enabled on startup or not at all. Also for switching from root to unprivileged user it needs user account preconfigured earlier in system. With systemd it can create unprivileged users on the fly without any administrator action in a transparent way for the system.

Using socket activation also enables for running server on demand when actual traffic is going to listening port.

Systemd integration provided for everyone secure, ready to use configuration which otherwise could take hours to recreate on its own and require technical knowledge to do so. Without it most users will end up with dnscrypt-proxy running as full root, possibly without keeping it updated.

Until today dnscrypt-proxy provided exceptional, modern and secure linux system integration even as it isn't included yet by most major distros, far beyond what projects like dnsmasq or unbound provide with full downstream distro support.

**EDIT:** If socket activation is such a problem (I don't know why but I happily look for compromises) then it can be easily dropped with minimal changes to systemd service which is still better than dropping everything we had.

jedisct1:
Having something that works exactly the same way across all operating systems would be way better.

Systemd as a way to start services is fine, but delegating application sockets to it causes a lot of headaches.

Instead of having everything in a single configuration file, now we have part of it in another unrelated configuration file. The example file doesn't work on Linux. Config files from other operating systems don't work on Linux. Instead of having a single service, now you need to start, stop, install, uninstall two of them on Linux. This is adding complexity for no practical benefits.

This creates a split. We need documentation for people using systemd sockets, and documentation for everybody else. This is bad.

And starting the server on-demand is a terrible idea for a DNS proxy. You want it to start as soon as possible, especially since the first thing it's gonna do is perform a benchmark across all servers and download the resolvers list.

Maybe things are better with dnscrypt-proxy 2, but with dnscrypt-proxy 1, the root cause of "it stops working after a while/after my laptop resumes from sleep/after my IP changes" was always systemd sockets.

Can I recycle old TCP connections to avoid DoS? If I control the sockets, sure. If systemd does, probably not. I don't know either if the systemd socket file can tune buffer sizes, attach epbf filters, and set other application-specific performance options. But even if it can, this is not the right place to do it, especially since people will mess with them (`TCP_NODELAY can be ignored` -- No, it shouldn't: performance will suck if support for pipelining is added later). If the application creates the sockets, this is not a problem any more, and everybody has the same guarantees. Linux/systemd users don't belong to a different class.

Dynamic users and other restrictions are also kinda scary. How does it play with cache files? Does it mean that these users, unlike everybody else, have to download the resolvers list every time they start the service?

Maybe not, I don't know, but what I do know is that this scenario is something that is not tested at all before new releases are tagged. On the other hand, if it works on any other environment, one can assume that it will also work everywhere else the exact same way. I'd like to avoid adding "except on Linux" to the previous sentence.

Systemd is complicated. The fact that each line of a basic configuration file apparently requires a comment to be understood confirms this. The systemd unit files are, with the WIndows batch files, the ones that received the most change requests, half of them introducing regressions. Everybody seems to want systemd, but nobody seems to be understanding it. So I'd rather use only a single service, that just starts/stop the thing, in the same context as every other operating system.

Switching to an unprivileged user would be great. But let's spend energy solving this for everyone. Not just for linux/systemd sockets activation. Do we want `chroot()`? Same thing, let's do it for everyone. The same way.

`dnscrypt-proxy -service install` apparently doesn't install a good enough systemd unit file to wait for the network to be available before starting the thing. Let's fix this, so that on Linux `dnscrypt-proxy -service install` works like everywhere else, and avoid fragmentation between Linux and the rest of the world once again.

ghost:
Thx for the detailed answer, I'll try to clarify some points.

> The example file doesn't work on Linux. Config files from other operating systems don't work on Linux. 

Traditionally on linux systems cache, logs and temporary files are put in their own directories in /var/cache, /var/log, /run, /tmp etc. Example config puts everything into same directory which is reasonable if someone installs it from github archive under /opt or other custom dir but not acceptable when used under package management. I'm sure if any distro will package dnscrypt-proxy they won't use example config without changes because of this same as [Archlinux already do](https://git.archlinux.org/svntogit/community.git/tree/trunk/configuration.diff?h=packages/dnscrypt-proxy).

> Instead of having a single service, now you need to start, stop, install, uninstall two of them on Linux. This is adding complexity for no practical benefits.

That isn't true. [Installing/Enabling](https://github.com/jedisct1/dnscrypt-proxy/commit/039479fa4bb55d5c8e7a3f9d939149960df06c54#diff-d2ed4509acf5f779cc0c185249fa98afL52) and [starting](https://github.com/jedisct1/dnscrypt-proxy/commit/039479fa4bb55d5c8e7a3f9d939149960df06c54#diff-d2ed4509acf5f779cc0c185249fa98afL9)  dnscrypt-proxy.service was enough thing to do. In that case dnscrypt-proxy.socket would be managed automatically through dependencies. For user it will be invisible.

> And starting the server on-demand is a terrible idea for a DNS proxy. You want it to start as soon as possible, especially since the first thing it's gonna do is perform a benchmark across all servers and download the resolvers list.

Fair point. Socket activation can be used for starting services on demand but it isn't mandatory. It's enough to start socket just before main service.

> TCP_NODELAY can be ignored -- No, it shouldn't: performance will suck if support for pipelining is added later

We may have misunderstanding here. Provided dnscrypt-proxy.socket was by default listening on both TCP and UDP sockets. It correctly applied TCP_NODELAY option to TCP socket and warn about failure when [it tried to apply it to UDP socket](https://github.com/jedisct1/dnscrypt-proxy/issues/288#issuecomment-383361003). That's why I wrote this can be ignored. It wasn't real issue and everything worked as desired.

> Dynamic users and other restrictions are also kinda scary. How does it play with cache files? Does it mean that these users, unlike everybody else, have to download the resolvers list every time they start the service?

No, there where [persistent directories](https://github.com/jedisct1/dnscrypt-proxy/commit/039479fa4bb55d5c8e7a3f9d939149960df06c54#diff-d2ed4509acf5f779cc0c185249fa98afL46) set for storage of cache and logs, created automatically. See Archlinux config mentioned above.

> Systemd is complicated. The fact that each line of a basic configuration file apparently requires a comment to be understood confirms this. The systemd unit files are, with the WIndows batch files, the ones that received the most change requests, half of them introducing regressions. Everybody seems to want systemd, but nobody seems to be understanding it.

I disagree. To make things work it needs couple of lines of config but people clearly wanted something better than that. It doesn't need commenting. I rarely see something commented in systemd files and I treat it as good sign that someone cared for explaining their choices. dnscrypt-proxy systemd files had enormous documentation which I myself contributed too and that was a sign that people really care of it and do their best to provide explanations to less informed people. They also tried to enhance them as much as it's possible using modern systemd options.

Unfortunately you may thought it's a sign of utter brokenness which make me sad when I think I lead to destroy something that I wanted to improve. Maybe there were misunderstandings and regressions but I'm sure the amount of scrutiny would make the final result better as some involved people had good systemd knowledge. Claiming that nobody understands systemd is a great exaggeration.

To be clear socket activation was made mandatory because it was seen as better choice in discussion about systemd enhancements  https://github.com/jedisct1/dnscrypt-proxy/pull/261#issuecomment-375961768 . There weren't raised any points about downsides of it before your above post. However it's possible to make it completely optional. I would volunteer for doing it if you shared above concerns about socket activation before deleting everything.

> dnscrypt-proxy -service install apparently doesn't install a good enough systemd unit file to wait for the network to be available before starting the thing.

I have to admit that I have no idea about configs on other systems than linux and don't know what `dnscrypt-proxy -service install` actually do. [This](https://github.com/jedisct1/dnscrypt-proxy/commit/039479fa4bb55d5c8e7a3f9d939149960df06c54#diff-d2ed4509acf5f779cc0c185249fa98afL11) (available in 2.0.11) should handle that situation (keep in mind that package from ubuntu ppa didn't update systemd files to latest versions and doesn't include network-online.target).

The unified behavior may be valuable for you as a developer but it also means we use on linux only 10% of what's possible because windows can't do something which makes users experience worse. All of this without touching the code which may badly affect other platforms. Using sockets or not I believe that systemd integration made linux the very first class platform capable of things which windows can't dream of.

Improving code security is valuable but orthogonal of what system integration does. I can have certain security guarantees without even looking at application code.

eli-schwartz:
@jedisct1 The number 1 reason to keep services here, is so Linux distros don't have to reimplement it downstream everywhere which leads to duplication. This is why the contrib folder is a pretty standard pattern in lots of open-source projects...

If you severely dislike socket activation, then how about systemd services which *aren't* socket activated, but still don't run as root because they use `CAP_NET_BIND_SERVICE` instead, which is a far more fine-grained approach to handing out privileges?

systemd services are not, contrary to popular belief, about setting tons of complicated security restrictions. That's just a completely optional bonus. At its heart it is just about telling the OS how you, the author, expect the program to be run... this is 100% compatible with running as root, depending on network-online, and having done with.

> Systemd is complicated. The fact that each line of a basic configuration file apparently requires a comment to be understood confirms this.

You mean https://github.com/jedisct1/dnscrypt-proxy/commit/dfdd469a814eaded8217c2e0d9264fb4bb4e8491 ???

99% of those comments are boringly pointless. I dispute the notion that, for example, a comment for network-online.target reading "wait for network" is useful documentation. You should just tell people to stop adding line noise.

ProtectHome=yes is not the same thing as ProtectHome=read-only (weaker), but the PR author who clearly did not read the `systemd.exec(5)` manpage didn't realize that and added *false* statements conflating the two.

"DNSCrypt-proxy provides and requires name resolution", yes, that's why the thing it is asking to come before is called "nss lookup", as in, name service resolution lookup. Real useful comment there.

"## Set O_NONBLOCK flag for socket unit"

Is that what "NonBlocking" does? How would I have ever guessed? Could you tell me where my power button is now?

"## Execute dnscrypt-proxy"

This comment appears next to the statement ExecStart=/usr/bin/dnscrypt-proxy which is clear evidence that the comment is wrong, as this is obviously the part of the service file that automatically generates pull requests on github containing selected extracts of the Gutenberg Bible. /sarcasm

"## Create symlink for systemd to pull in the unit when starting multi-user.target"
"## Can be found in /etc/systemd/system/multi-user.target.wants/"

Why is it necessary to add comments describing low-level boilerplate about the actual systemd service file sytax? Is the next "documentation" PR going to add a comment to the "[Service]" line, explaining how this is systemd syntax for beginning a systemd service file?

#395 by @brainscar should have been rejected outright. It added a great deal of false information, useless duplication of systemd documentation and in some cases useless duplication of basic reading comprehension.

It did one thing right, in adding Wants=network-online.target, because that goes together with After. But then it added a documentation commentary claiming that this "requires" using some *-wait-online.service. This is completely and utterly wrong; the purpose of the target is to allow users to *opt in* to some wait-online service, and autodetect other services that want to order themselves after that...

This all worked fine as contributed by @dvzrv **who is a linux distro developer and understands this stuff**, and should therefore be trusted to set reasonable, sane configs.

#389 changed it to depend on network-online, but forgot to add the complete stanza, which could've been trivially fixed. It also doesn't really hurt to be incomplete.

Please add these systemd files back, but remove the so-called documentation, which:

- does not help people who don't understand how systemd services work and are simply trusting their distro maintainers

- actively hinders people who understand how systemd services work, as they now have to filter through far more comments than actual configuration options, and cannot see the forest for the trees

This is like accepting a pull request for https://github.com/jedisct1/dnscrypt-proxy/blob/084bfe67fd7a4983c85757439c8f9290a7878fc6/dnscrypt-proxy/main.go#L14-L16 which adds the "documentation":

```
# AppVersion is the current version of the program
```

jedisct1:
Maybe add a "guidelines for people building LInux packages" to the Wiki?

ghost:
> It did one thing right, in adding Wants=network-online.target, because that goes together with After. But then it added a documentation commentary claiming that this "requires" using some *-wait-online.service. This is completely and utterly wrong; the purpose of the target is to allow users to opt in to some wait-online service, and autodetect other services that want to order themselves after that...

Let me quote [official systemd documentation](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/):

> The right "wait" service must be enabled too (NetworkManager-wait-online.service if NetworkManager is used to configure the network, systemd-networkd-wait-online.service if systemd-networkd is used, etc.)

Without `*-wait-online.service`, `network-online.target` has completely no effect and dnscrypt-proxy will try connecting to dns servers before network connection is up which means that comment was totally legit. Actually it was almost copy-pasted from official documentation. Pleas refrain from making bold claims about something being "completely and utterly wrong" without double-checking the facts unless you want to convince @jedisct1 that "nobody understands systemd" is indeed a true statement.

eli-schwartz:
> will try connecting to dns servers before network connection is up which means that comment was totally legit

Just like every other system service which depends on a network connection, dnscrypt-proxy will get **no benefit** out of specifying network-online if there is no network management utility-specific service configured and active. This is hardly a systemd requirement; if it was, the service itself would depend on NetworkManager/networkd services. The whole point of this is it is meant to gracefully degrade to the equivalent of network.target if it isn't actually hooked into some wait-online service. That is entirely under the user's control, and some linux distros will autoconfigure this for you while others expect you to [read the setup guide](https://wiki.archlinux.org/index.php/NetworkManager#Enable_NetworkManager_Wait_Online) and make a personal choice.

Ominous warnings serve only to... well, get the project developer to worry about the seemingly confusing mess of decidedly not out-of-the-box instructions and lengthy manuals. Which is silly because no one is actually using those instructions.

> Actually it was almost copy-pasted from official documentation.

That's sort of the whole problem. The so-called documentation in this service file is copy-pasted without important context, from places where it is described far, far better. If there was any legitimate reason to litter comments all over this service file, it would have made a lot more sense to **just** link that url.

`systemctl help network-online.target` ==> systemd.special(7) ==> describes the purpose of this unit, linking to that wiki page.


This commentary litter is exceedingly uncommon as far as projects shipping default systemd configs go.

ghost:
The problem is dnscrypt-proxy will print errors and people [will open issues about that](https://github.com/jedisct1/dnscrypt-proxy/issues/398). There was `TCP_NODELAY` thing which confused many people including main developer. Too little or too much documentation, someone will always find a reason to complain.

Anyway it seems systemd is gone for good so perhaps this discussion doesn't lead anywhere.

dvzrv:
@jedisct1 I can only rehash what @eli-schwartz wrote on the topic: Please don't remove the systemd unit files, just because some people want to over-document them.
I have spent some time doing pull requests for this and getting the current version of dnscrypt-proxy updated in the Arch Linux repositories.
Would be great if that was not in vain.

jedisct1:
These files should be part of the documentation. A dedicated page about starting the server using systemd would be fantastic.

Anybody could easily improve it, add distribution-specific things, document every single line as much as they want, add links to the official systemd documentation, and more.

Having these example files in the source code doesn't really make sense. And if we go that route, we need to add scripts and configuration files for all operating systems. Which would be just as useless, because this is not what official packages for these operating systems will be using anyway.

For people who are not package builders, and don't want to use packages, and just want to download files from here, the `-service install` function could be improved. Especially for systemd, so that it installs the right thing for each distribution. *that* could really get some help from all the systemd experts we got here.

dvzrv:
> These files should be part of the documentation. A dedicated page about starting the server using systemd would be fantastic.

No. They should be shipped and reusable downstream.

> And if we go that route, we need to add scripts and configuration files for all operating systems. Which would be just as useless, because this is not what official packages for these operating systems will be using anyway.

I'm using the systemd units in the Arch Linux package. So no, they're not useless, but vendor defaults, that are installed to `/usr/lib/systemd/system` or `/lib/systemd/system` (depending on your Linux distribution).
[These are perfectly fine then](https://github.com/jedisct1/dnscrypt-proxy/tree/master/windows) or will they also be randomly deleted?

> For people who are not package builders, and don't want to use packages, and just want to download files from here, the `-service install` function could be improved

Just to get this right: You want to replace plain text files with a builtin command, that generates a service file for the user/ or system, but will require root to install it to `/etc/systemd/system`? What is the upside to maintaining all of that logic (this time for real in your source code)? I fail to see any advantage.
Files below `/etc/systemd/system` are meant for the administrator of the machine, overriding the vendor defaults.

I'm just trying to understand the rationale behind all this, as packaging this software is not necessarily straight forward. I have to patch dnscrypt-proxy.toml for system-wide use, make it FHS compliant and now also supply unit files...
Rage-quitting version 1.x doesn't give me the confidence this is going to be the only change like this.

eclipseo:
As a Fedora packager, I'd like very much if you kept the service file in the repo, I'm using them with no issue on Fedora 28. I don't really want to reimplement them downstream, it would be such a hassle and would lack consistency with what other distributions are doing.

eli-schwartz:
> For people who are not package builders, and don't want to use packages, and just want to download files from here, the -service install function could be improved. Especially for systemd, so that it installs the right thing for each distribution. that could really get some help from all the systemd experts we got here.

I don't feel confident writing golang at the best of times, and I'm not sure how hardcoding a systemd service file as a string literal inside some go file would be a technical advantage over providing the file itself.

> Having these example files in the source code doesn't really make sense. And if we go that route, we need to add scripts and configuration files for all operating systems. Which would be just as useless, because this is not what official packages for these operating systems will be using anyway.

Why not? Why is Windows special, and why do you think any community-contributed systemd services would be submitted in a non-working state? If you're willing to work on systemd-managed files from within some autogenerator, why couldn't the same static file be provided on its own? You could then compile it into the binary and make -service install just install those files.

As far as I can tell, the only actual issues with using the provided systemd service (after minimal modifications to use the downstream filepaths rather than the ones which assume /opt) turned out to be completely bogus. Even with systemd-managed sockets.

If people have problems, do like developers usually do, and 

- ask the user to try running the software bare, without intervening layers like systemd,
- if that works, they should consult their local systemd experts, which usually means distro support channels, 
- wait for knowledgeable people to submit changes to the service file. Don't just accept anything, ask them to explain what it all means. If they cannot do that well enough for you to understand without lots of Linux expertise which you don't have, they might not be sure of what they're doing either....  Also, work with distro packagers as they have a vested interest in getting this to work as well as possible in a distro-agnostic manner. (I suspect @dvzrv and @eclipseo would be happy to be @mentioned for proposed changes to it.)

jedisct1:
Here's a concrete plan:

- Implement application-defined templates, including external templates in https://github.com/kardianos/service.
- Bring back an example, as a template.
- Implement a "must run after the network has been initialized" property, usable within that template.

eli-schwartz:
From their README:

> Dependencies field is not implemented for Linux systems and Launchd.

That makes it difficult to implement services which require dependencies. I'm not sure how to reimplement an entire thirdparty multiplatform support library just to hardcode non-generic arguments into a platform-specific, project-specific thing. Maybe if they supported this upstream, I could suggest the right things to set...

On the plus side, this never worked with -service install either, so dropping this crude thirdparty dependency can only improve things. I'm not sure what the precise utility of this external templates thing is, since:
- AFAICT from looking at the *.go files with their string literals, it only properly supports 1/3 of the supported operating systems, which is to say 1/5 of the supported service managers,
- It would need some way of generalizing "what is a network connection" across multiple OSes, which is a lot harder than doing the one thing which this template seems to do -- that being "execve(2) a commandline using a given UID and track its PID to see when it dies"... **unless you use Windows**, in which case there is a *4th*-party project which implements all the actual service creation and supported parameters, and which I cannot understand the featureset of. It simply doesn't let you do anything complicated, and it's not meant to, because complicated things like "depend on a network manager", or "run with a timer on certain calendar events" or "create a socket to listen on", or, well, anything more complicated than execve(2) is not easy to generalize.
- Despite existing for 6 years, it's never implemented the functionality for anything other than Windows, and even that functionality was from kardianos/service#3 which essentially says that since the *4th*-party project which implements actual Windows services supports the dependency variable, kardianos/service should support blindly forwarding the content to said 4th-party project. This is hardly a ringing endorsement, and gives me little hope that anyone else thinks it can be implemented. It's less of a bug that Linux/macOS don't support it, and more of an easter egg that Windows does. I guess it works out for Windows, since it only works for one platform, you can hardcode the Windows name in every project using these templates, and no else one will notice the difference since it was never supported to begin with.
- Or not. This supports Windows first, and everything else as an afterthought. So if something isn't supported on any other platform, they can just say "oh, it's a bug that that doesn't work", except really it doesn't work because no one ever really tried to support anything other than Windows or "I give up, let's just execve(2) and say that is good enough". Because doing more than that is actually *less* work than simply doing it by hand.
- I think the real reason it requires so many libraries to do this on Windows, is because you cannot just have a file describing the options with simple key-value pairs, so you need some huge generator instead.

I'd rather not learn a new programming language from scratch, in order to contribute code to some project that will most likely not even work because you'd need to #ifdef every single operating system anyway to describe the unique operating-system-specific name for the network manager (or hardcode it upstream in some giant list which requires waiting on upstream to support each and every dependency by name), which doesn't either help the case where you want to tell the service that it doesn't need or use various other concepts like the ability to read user directories or modify the system's long-term files... something which would generally be functionality specific to Linux/systemd.

So even if the template supported dependencies, we'd still need to do the same dnscrypt-proxy specific work, resulting in something maintained partially in one place and partially in another place.

Given that implementing anything meaningful that has operating-system-specific behavior (unlike "execve(2) this program") will almost certainly require every single project to implement things themselves rather than utilizing this pot of gold at the end of the rainbow... I really, really don't get the point. Just add a linux/ directory next to the windows/ directory, and work with the operating system experts who want to help you.

You'd end up hardcoding exactly all the same operating-system-specific key-value pairs, except in a systemd source file instead of a golang source file.

...

Or I could be proven wrong and someone else will come up with some brilliant way to template the untemplateable.

One partial workaround would be if it supported arbitrary footer text, so you could implement the bits kardianos/service does not, via extra Linux-specific arguments. But that still doesn't seem like it would support anything other than .service files, so no .socket files and no other interesting things like .timer files, .path files for path-based activation, .mount files for depending on a filesystem mount, etc.

jedisct1:
Too long, didn't read.

The followup comments should be on a pull request.

jedisct1:
Closing, in order to avoid breaking down and completely removing Linux support, since doing "echo 'app &' >> /etc/rc.local" has become rocket science.

Issue #458, opened 2018-05-20
Feature(for release) request: statically linked release


honwen:
Release in https://github.com/jedisct1/dnscrypt-proxy/releases

Some are ```statically linked```, like ```linux_mips```
Some are ```dynamically linked```, like ```linux_x86_64```

Can U release binary all statically linked, please? THX a lot.

jedisct1:
Duh?

Go cannot produce dynamically linked executables.

jedisct1:
Ah, interesting, when the thing is compiled on Linux (as Travis does), it depends on `/lib64/ld-linux-x86-64.so.2`, but if you compile it on macOS, you get a static executable.

Apparently, the `-tags netgo` can avoid this.

But you obviously have far more experience with Go than me, so I'd gladly accept a pull request to change the Travis file and build static files, provided that it doesn't remove features or degrades performance.

Thanks for your help!

PR #459, opened 2018-05-20
Ensure static linked binary (for linux)

honwen:
Refer: https://github.com/jedisct1/dnscrypt-proxy/issues/458

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,49 +75,49 @@ script:
 - ln ../LICENSE example-dnscrypt-proxy.toml example-*.txt netbsd-amd64/
 - tar czpvf dnscrypt-proxy-netbsd_amd64-${TRAVIS_TAG:-dev}.tar.gz netbsd-amd64
 - go clean
-- env GOOS=linux GOARCH=386 go build -ldflags="-s -w"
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks!

What is the difference between `-tags netgo` and `CGO_ENABLED=0`?

`CGO_ENABLED` produces a smaller binary. Does it disable features/reduce performance?

honwen:
About ```-tags netgo```, Refer: https://github.com/docker-library/golang/issues/152
> Try adding -tags netgo -installsuffix netgo (which forces the all-Go networking implementation and should thus remove the cgo dependence).

About ```cgo```, Refer: https://golang.org/cmd/cgo/
In my experience, some go lib implement build with cgo have better performance.
For example, with ```CGO_ENABLED=1```, resolver (in ```netgo```) will be work with ```libc```, and work with system level cache (means better performance). With ```CGO_ENABLED=0```, resolver  will work its own(without cache). 
Nowsadys most Linux OS is built with ```glibc```, but some with ```musl```(glibc-linked binary cant work), like ```alpine```, ```openwrt``` and some on. More info: http://www.etalabs.net/compare_libcs.html.

```CGO_ENABLED=1``` will be default, when ```GOOS``` is the same as build system.

Issue #441, opened 2018-05-10
Feature request: add the ability to drop privileges on Unix


jedisct1:
fork+exec seems to work fine on Linux/BSD/macOS. We can create sockets, and easily pass them to children: https://gist.github.com/jedisct1/92a75f1dc7a4c2c1cad4802131994736

Feature request: implement this in `kardianos/service`.

jedisct1:
There are still some issues, such as the root process that remains around (or else, it breaks installation as a service).

But we can take advantage of this to implement a simple supervisor, that restarts the child if it ever dies.

publicarray:
> There are still some issues, such as the root process that remains around (or else, it breaks installation as a service).

I was afraid that there might me something that would break 😢

Regarding the supervisor, It's a good idea but I also think that this stuff adds code for not much gain. Many init systems can already be configured to restart a service if it dies.

PR #467, opened 2018-05-25
Drop privileges with exec

publicarray:
@jedisct1 What do you think? I know its a bit messy, but I like to get some feedback first. I'm using your method from #441 

From the issue I take it that would have preferred this to be implemented in `kardianos/service` with an API instead?


publicarray:
Please don't merge as this does not work on Windows

jedisct1:
Still, this is fantastic, thanks a lot!

Having this implemented in `kardianos/service` (or equivalent) would have been neat to avoid reimplementing the same thing in every single application. But having this already fixed in dnscrypt-proxy is already awesome.

Wondering how this affects logging, though. We probably need to do the descriptor passing thing with logfiles descriptors, too.

publicarray:
Yea totally missed the log file descriptor. How would you feel with something like this?

```go
if !proxy.child {
	fds = append(fds, dlog.GetFileDescriptor())
} else {
	dlog.SetFileDescriptor(os.NewFile(uintptr(3), "logFile"))
}
```

I took a quick look at kardianos/service. I think because the package is just just concerned with creating services I can't help thinking of the gnu model (do one thing only but do it well). It might be better to create a brand new package just for dropping the privilege with exec. Basically I'm not confident enough to do it in  kardianos/service.

publicarray:
Unfortunately I have no idea what to do with the logging plugins

jedisct1:
Log files need to be owned by the proxy user, since they are automatically rotated. So, I don't think anything has to be done.

publicarray:
Ah that makes sense. I was over thinking it.

PR #471, opened 2018-05-28
Asterisk Pattern

catinred2:


jedisct1:
I don't understand. What does this do?

catinred2:
"apple.com.akadns.net" would match "a.apple.com.akadns.net" , but under some circumstance , there is no easy way to put forwarding rule for "store-010-lb.blobstore-apple.com.akadns.net". 
It is only my personal need.  close it if you think it not necessary.

jedisct1:
Mmmm... yes, that looks more like a hack for a peculiar need.

If we want to extend this, it would be probably better to use the `PatternMatcher` struct, as done for blacklisting, whitelisting, cloaking, etc.

PR #492, opened 2018-06-13
Fix systemd socket connections

### LINKED ISSUE #480: systemd UDP socket missing in version 2.0.15
I upgraded from 2.0.14 to 2.0.15 on my Raspberry Pi. According to syslog, everything is running OK, but DNS requests do not get forwarded anymore: if I do a nslookup, the request gets a timeout. Reverting back to 2.0.14 makes everything work normally again. 

I do not know which debug info you need, please let me know how to support.

dwimmer:
Upstream systemd go library broke use of TCP and UDP sockets at the same time.
Changed to use lower level API to work around this. Also improved logging of
systemd socket connections to include systemd unit file name and address.

This should fix  #480. Ideally the systemd library would be improved to allow calling both activation.Listeners() and activation.PacketConns() again so we don't have to duplicate as much code.

### DIFF
--- a/dnscrypt-proxy/systemd_linux.go
+++ b/dnscrypt-proxy/systemd_linux.go
@@ -9,24 +9,19 @@ import (
 )
 
 func (proxy *Proxy) SystemDListeners() error {
-	listeners, err := activation.Listeners()
-	if err == nil && len(listeners) > 0 {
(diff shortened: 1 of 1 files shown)


jedisct1:
Awesome, thank you!

Issue #448, opened 2018-05-17
Feature request: DNS Cache statistics


scubasunny:
Hi,

would be nice to get some statistics about the cache efficiency, cache hits and cache misses - also if cache is filled or has still some space.
could be useful to configure cache size properly

jedisct1:
+1

JasonSwindle:
+1 as I have this same request:

https://github.com/jedisct1/dnscrypt-proxy/issues/337

meguroyama:
May need more research but I'm looking into see if we can use [expvar ](https://golang.org/pkg/expvar/)to monitor dnscrypt in [netdata](https://github.com/firehol/netdata/wiki/Monitoring-Go-Applications).

IceCodeNew:
+1

Misiek304:
+1

beelze:
+1 for cache stat and cache memory usage

mikegleasonjr:
A Prometheus exporter or Prometheus-compatible endpoint would be great!

jedisct1:
Yes, using Prometheus instead of some ad-hoc protocol would be great.

But it looks like you have far more experience with Go that I do, so go ahead :)

mikegleasonjr:
@jedisct1 I will consider it for sure... thanks for the project. Running it in a docker environment easily with pi-hole!

J0s3f:
Would be great to not only have cache statistics but also statistics about the DNS servers.

scubasunny:
@jedisct1 : Hi, some reason why you closed this request ?

Thx

PR #493, opened 2018-06-14
Allow plugins to report their status

mitchelhaan:
This is a proposal to implement #448 with potential for other plugins to provide status as well.

Main changes:
- Each plugin now includes a Status() function
- New config field to enable statistics collection
- Example implementation of status for the cache plugin, reporting cache statistics

Cache statistics:
- entries: number of DNS entries stored in the cache (which could include expired entries)
- hits: queries that were resolved from the cache
- misses: queries that needed to be forwarded
- expirations: queries that were cached but the TTL expired

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 # ![dnscrypt-proxy 2](https://raw.github.com/jedisct1/dnscrypt-proxy/master/logo.png?2)
 
-A flexible DNS proxy, with support for modern encrypted DNS protocols such as [DNSCrypt v2](https://dnscrypt.info/protocol) and [DNS-over-HTTP](https://tools.ietf.org/html/draft-ietf-doh-dns-over-https-10).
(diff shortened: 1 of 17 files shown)


jedisct1:
Hi!

And thanks!

This is a good start, but there are a few issues with this approach:

- It introduces some overhead (especially the mutex) to everyone, even when statistics are not enabled.
- Periodically writing unstructured data to a log file may be useful for debugging, not so much for real applications. We can't use this for a web UI, for example. A gRPC interface, chunked HTTP responses, or a standard Prometheus API may be better an option.

mitchelhaan:
Yeah, that makes sense. I was mainly concerned with collecting the data and hadn't come up with a good idea for presenting it yet.
Adding a field for enabling/disabling statistics would be easy enough though.

jedisct1:
I'm planning to write a Pi-Hole replacement. A web UI that shows stats, logs, and let you edit the configuration.

So anything that can be updated live and is easy to parse would be preferable.

mitchelhaan:
I'm thinking about an API structure like so:

Global proxy config/info:
`/api/v1/proxy/{status,config,reload?}`

Plugin-specific:
`/api/v1/plugin/<name>/{status,config,reload?}`

So a `GET` request for `/api/v1/plugin/cache/status` would return a JSON object containing the cache plugin status (enabled/disabled, statistics, etc)

PR #511, opened 2018-06-19
Quick typo fix in example config.

johnspurlock:


jedisct1:
Good catch, thank you!

PR #524, opened 2018-06-23
Config: Add option to specify count of fastest servers to pick from.

ssteinberg:
Removes LBStrategyP2 and adds a new config option 'lb_strategy_fastest_count'. When used in conjunction with 'lb_strategy=fastest' allows to pick the count of fastest servers to use.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	CertIgnoreTimestamp      bool   `toml:"cert_ignore_timestamp"`
 	EphemeralKeys            bool   `toml:"dnscrypt_ephemeral_keys"`
 	LBStrategy               string `toml:"lb_strategy"`
+	LBStrategyFastestCount   int    `toml:"lb_strategy_fastest_count"`
 	BlockIPv6                bool   `toml:"block_ipv6"`
 	Cache                    bool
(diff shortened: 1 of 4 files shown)


jedisct1:
Why?

Besides being a breaking change, when is this an improvement over the well-studied [power of two choices](https://www.eecs.harvard.edu/~michaelm/postscripts/mythesis.pdf)?

I'm not a big fan of adding `lb_strategy_fastest_count` because it only applies to one specific strategy. It doesn't make sense with randomized LB, or additional strategies.

ssteinberg:
So does ph. I find the p2/ph too limiting, it does make sense to select more servers and not only from a load-balancing point of view.
In addition, the current LBStrategyPH selection seems off, it is identical to P2.

This is just a suggestion and needs not be breaking. I don't like the additional config either, ideally it would be an option to select a fraction of total servers or an absolute count.

jedisct1:
`ph` was just an experiment, but I doubt it ever was useful.

It is as clunky has the "N fastest servers". Even if a specific `N` value seems to empirically be optimal at some point in time, it won't be the case any more shortly after, if only because servers come and go. 

The load balancing algorithms could be improved a lot. In particular, we don't keep track of the number of inflight queries for each server yet. This is a bandit problem, with many possible approaches to get closer to the optimal parameters. But adding more knobs for users to turn is not one of them.

PR #538, opened 2018-07-04
Add dnscrypt-proxy vector logo

lukateras:
I redrew dnscrypt-proxy logo in Inkscape. Tell if I got any details wrong.

See preview: https://github.com/yegortimoshenko/dnscrypt-proxy/blob/svg-logo/logo.svg


publicarray:
Good work! A fine detail but the front right leaf doesn't quite have the right colour, it doesn't quite pop with the red. It's almost too brown. If you want to be picky I also think the green oval has a slight glow on it in the original.

Come to think of it did you use the correct colour profile? I think SVG assumes sRGB by default

PR #559, opened 2018-08-03
Minor improvements for the user_name option

publicarray:
I'm not sure if [kill](https://golang.org/pkg/os/#Process.Kill)ing the process is the best way to do it. Maybe we need to investigate if [cmd.Process.Wait()](https://golang.org/pkg/os/#Process.Wait) or [cmd.Wait()](https://golang.org/pkg/os/exec/#Cmd.Wait) is a better option.

### DIFF
--- a/dnscrypt-proxy/main.go
+++ b/dnscrypt-proxy/main.go
@@ -98,8 +98,8 @@ func (app *App) Start(service service.Service) error {
 }
 
 func (app *App) AppMain(proxy *Proxy) {
-	proxy.StartProxy()
 	pidfile.Write()
+	proxy.StartProxy()
 	<-app.quit
 	dlog.Notice("Quit signal received...")
 	app.wg.Done()
--- a/dnscrypt-proxy/main.go
+++ b/dnscrypt-proxy/main.go
@@ -110,6 +110,7 @@ func (app *App) Stop(service service.Service) error {
 	if pidFilePath := pidfile.GetPidfilePath(); len(pidFilePath) > 1 {
 		os.Remove(pidFilePath)
 	}
+	killChild()
 	dlog.Notice("Stopped.")
 	return nil
 }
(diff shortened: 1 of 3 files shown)


jedisct1:
That looks great! Thank you so much, Sebastian!

publicarray:
Hi Frank! Do you think this patch could warrant a new release ?

publicarray:
@jedisct1 sorry to ping you ^

jedisct1:
Just wanted to check this on a variety of operating systems before tagging a new release :)

publicarray:
No worries. It's always good to test first 👍
It's just that I want to stop the zombie (process) apocalypse before it becomes a problem 🙈

jedisct1:
It doesn't seem to work on OpenBSD :( The zombie is still around.

publicarray:
From my short investigation It appears that the Stop() function is never called on OpenBSD. https://github.com/jedisct1/dnscrypt-proxy/blob/33bcff7d4a923474e0051f8dd2b946ba88404f5f/dnscrypt-proxy/main.go#L109

publicarray:
Yea I was relying on this piece of code to be present on all platforms.

https://github.com/kardianos/service/blob/7a88211485df5d3da6ec68f4f8a5ff9bec297327/service_darwin.go#L199-L206

```go
s.Option.funcSingle(optionRunWait, func() {
		var sigChan = make(chan os.Signal, 3)
		signal.Notify(sigChan, syscall.SIGTERM, os.Interrupt)
		<-sigChan
	})()

return s.i.Stop(s)
```

publicarray:
@jedisct1 How would you like to proceed? we can open an issue with kardianos/service whether they would implement the signal to call the Stop() function in OpenBSD or we can call Stop() ourselves when SIGTERM is signalled.

jedisct1:
Maybe do both, as `kardanios/service` doesn't seem to be frequently updated :(

PR #570, opened 2018-08-15
a quick fix to dnscrypt-proxy old pledge call

eau-u4f:
pledge(2) in 6.3 evolved as unveil(2) is coming in 6.4
as it was killed during the attempt to run it, quick investigation showed that pledge call was the old prototype of the (6.2) pledge(2) call.
In addition the proper restrictions have not been completely investigated but were missing "unix" pledge since the hashicorp gsyslog wrapper code is NOT using log/syslog but implement a socket()/connect() call to a "unix" socket... hence the pledge update
the author probably know better all the intricacies required, it seems to run properly now.
did not change the building travis-ci stuff, I'm not using it extensively.
please let me know if this is correct, hope this helps.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -193,9 +193,9 @@ func ConfigLoad(proxy *Proxy, svcFlag *string) error {
 
 	flag.Parse()
 
-	if *child {
-		PledgeChild()
-	}
+	//	if *child {
+	////		PledgeChild()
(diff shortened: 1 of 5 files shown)


jedisct1:
Golang not using the libc leads to that kind of surprise when the syscall ABI changes.

Let's completely remove `pledge()` support for now. We'll revisit this after the next release of the compiler.

Issue #608, opened 2018-10-09
2.0.17 still doesn't work on Android SDK 19


sfionov:
```
shell@lt01wifi:/data/local/tmp $ ./dnscrypt-proxy                              
CANNOT LINK EXECUTABLE: cannot locate symbol "sigfillset" referenced by "./dnscrypt-proxy"...
1|shell@lt01wifi:/data/local/tmp $ getprop ro.build.version.sdk                
19
```

sfionov:
Ah, sorry, I see that there is commit https://github.com/jedisct1/dnscrypt-proxy/commit/c53b28d1f1093f1dccc372cd2e4462e5a377d712 that removed -D__ANDROID_API__=19 flag again.

Can you please add it for armeabi arch to support older devices?
Also seems that arm64 arch builds only with -D__ANDROID_API__=21 because it didn't exist earlier.

jedisct1:
It doesn't compile any more if I add it back. Maybe this is not the way to add it?

Since you obviously know Android way better than I do, can you submit a pull request with the required changes to the `travis.yml` file?

Thank you!

sfionov:
I've created a PR.

jedisct1:
Thank you!

PR #610, opened 2018-10-10
Use Android NDK r18, use API 19 for 32-bit Android builds

sfionov:
Older androids had some function like `sigfillset` implemented as macros, and `android-19` is 
the last backward-compatible and forward-compatible version of unified headers.

https://github.com/jedisct1/dnscrypt-proxy/issues/608


jedisct1:
Merged, thank you!

PR #655, opened 2018-11-28
Removed securemecca.com as it has expired and does not currently host…

enxoco:
Securemecca.com domain has expired and does not currently host any blacklist content.  As such it should be removed from config files.


jedisct1:
Goodbye, Securemecca.

Thanks for the heads up!

mibere:
Is there a reason why with PR was closed, instead of merged? :)

PR #697, opened 2019-01-16
Minor domains-blacklist fixes

losuler:
## Changes

I noticed that `https` was on a few URLs in `domains-blacklist.conf` that weren't in `domains-blacklist-all.conf` and one that wasn't in both (`www.malwaredomainlist.com`).

I also noticed there was the characters `.\` in the `hosts-file.net` URLs, which I assume was just a typo.

## Notes

I tested each URL in browser and then with `./generate-domains-blacklist.py -c domains-blacklist-all.conf` which didn't seem to result in any errors.


jedisct1:
Thank you!

PR #742, opened 2019-03-02
logging enhancement CLOAK

corsmith:
Currently responses that have been modified by plugins_cloak show up as PASS in the log.  Create a new PluginsReturnCodeCloak so cloaked responses are logged as CLOAK


jedisct1:
This is cool, thank you!

Maybe we should have the same for forwarded queries?

PR #747, opened 2019-03-06
Add Matrix Chat Badge

CHEF-KOCH:
* Add Matrix.org Chat badge
* Place the Travis build status badge together with the matrix chat badge under the main logo. Reason? In case more badges are coming (_you never know_).


jedisct1:
Thanks!

PR #748, opened 2019-03-06
Center main logo

CHEF-KOCH:


jedisct1:
Not a big fan of having HTML in the Markdown file and hard-coded sizes.

CHEF-KOCH:
Okay, so be it. I close because the other option would be CSS which is even more uglier.

PR #764, opened 2019-03-19
Replace “find” with “findstr” in batches

sergeevabc:
DNSCrypt-Proxy 2.0.8 was the last release which didn’t use “find”, then it was introduced. And up to this day (2.0.21) I had to replace “find” with “findstr” in batches, because “find” doesn’t work when codepage is 65001 (Unicode). The proposed change is intended to correct this misunderstanding once and for all.

### DIFF
--- a/windows/service-install.bat
+++ b/windows/service-install.bat
@@ -3,7 +3,7 @@
 setlocal EnableExtensions
 title DNSCrypt-Proxy
 
-whoami /groups | find "S-1-16-12288" >nul && goto :admin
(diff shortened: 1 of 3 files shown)


sergeevabc:
Also consider changing extensions from ```bat``` to ```cmd``` to reflect these scripts are meant for the newer Windows versions (in our case, after Go internal changes in DNSCrypt-proxy [2.0.17][1], Windows 7 is the required minimum).

[1]: https://github.com/jedisct1/dnscrypt-proxy/releases/tag/2.0.17

jedisct1:
Thank you!

Issue #714, opened 2019-02-01
Watch blacklist file to reload them automatically


Belphemur:
Hello,

Would it be possible to implement a watcher for the blacklist file ?

You have the [fsnotify](https://github.com/fsnotify/fsnotify) lib that seems to be quite easy to add to the project.

This way, you could watch for file change for the blacklist and simply reload them.

Because even now, with the new code, it's still slow when you have more than 500k entries. 

My use case is, I'm replacing pihole-FTL by dnscrypt-proxy since it's way more optimised in the term of memory usage for the same dataset of blacklist.

I'm also providing the service publicly and I don't really want to have any dns outage the time to restart the service for blacklist update.

ps: fsnotify also works on Windows.

jedisct1:
Watching for file changes is trivial, but gracefully reloading configuration files is always a tricky thing to do.

You may have noticed a `Reload()` function in the plugin interface. This is exactly what it's here for. But currently, `Reload()` functions are empty.

It may do something like spawn a new goroutine that builds a new instance of the plugin, and then atomically replaces the previous one with the new one. And there's probably a generic way to do it for any plugin instead of reimplementing the same thing for every single plugin loading a configuration file.

It's been on my TODO list since the beginning, but I haven't found the time to do it. Could you possibly help?

Belphemur:
Good point, I see it's not going to be the easiest system to implement, especially the logic behind each plugin for a graceful reload.

Another way to do it, supporting SIGHUP using [TableFlip](https://github.com/cloudflare/tableflip) of CloudFlare. They made a [whole article](https://blog.cloudflare.com/graceful-upgrades-in-go/) about their lib.

This way, the whole process is reloaded, all the config. They based their lib on how NGINX is doing the reload.

Eidt: I admit I have no idea how it would work with systemd socket.
And obviously the address bindings would still need a restart.

jedisct1:
Tableflip also seems to require TCP connections. But DNS mostly uses UDP; there are no connections.
Also, it is only designed for Linux, whereas dnscrypt-proxy has to support many more operating systems, including Windows.

Belphemur:
TableFlip support any socket that is provided by the network package of golang. Hence, UDP also supported. The idea been TableFlip is the main thread taking care of the network.

[...]

PR #780, opened 2019-04-06
Watch plugin config files, automatic reload plugin on a file update

jrnewell:
Made an attempt to add the requested feature outlined in this issue (https://github.com/jedisct1/dnscrypt-proxy/issues/714)

- Watch config files (`'forwarding-rules.txt' 'cloaking-rules.txt', ...`) of enabled plugins using https://github.com/fsnotify/fsnotify.  If the config files are updated, the plugin is rebuilt in the background and once initialized replaces the old plugin.
- If the plugin has a logger (i.e. `ip-block`), the logger is migrated to the new plugin instance
- Add a 2 second debounce, so if many updates occur simultaneously (and to handle the issue where MacOS can get dual update notifications from SpotLight), the plugin will be recreated only once
- Add a config option `hot_reload_plugins` to enable this feature (off by default).

Let me know what you think.  Go is not my primary language :)

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -82,6 +82,7 @@ type Config struct {
 	HTTPProxyURL             string                     `toml:"http_proxy"`
 	RefusedCodeInResponses   bool                       `toml:"refused_code_in_responses"`
 	BlockedQueryResponse     string                     `toml:"blocked_query_response"`
+	HotReloadPlugins         bool                       `toml:"hot_reload_plugins"`
 }
 
 func newConfig() Config {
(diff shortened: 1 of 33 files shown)


ghost (approved):

publicarray:
Ping @jedisct1 

@jrnewell I think this is a great idea!

jedisct1:
This is an excellent idea, and the proposed changes are excellent. And thanks for thinking about the debounce feature!

Reloading the configuration was originally planned using the `Reload()` trait that all plugins have to implement already. I have some early code using this, and after seeing your PR, my plan was so do a mix of both approaches. But I still have to find some time to do it.

unixfox:
why closing this PR? this feature would be awesome when managing a dynamic blacklist with DNSCrypt-proxy

jedisct1:
That feature will be implemented differently.

PR #783, opened 2019-04-12
Update fallback_resolver to 8.8.8.8

bagusnl:
9.9.9.9 didnt work, atleast in my country. 8.8.8.8 is more stable I guess.


mibere:
I'm against the Google server as default fallback resolver. Default should keep Quad9.

Mikaela:
Does it have to be the primary resolver with easy to remember numbers if no one is going to type it anyway? How about `149.112.112.112` which is the secondary DNS of Quad9?

jedisct1:
Some people may indeed not be comfortable with Google.

We can use `149.112.112.112` but it's not obvious that this is Quad9 and some people may find this IP address suspicious.

I'm not sure that we should change this, especially since `9.9.9.9` being blocked doesn't seem to be very common and I've never heard of it being blocked in a whole country, especially with `8.8.8.8` working in that same country. It rather looks like a specific network configuration issue.

PR #791, opened 2019-04-14
 cache plugin: Fix caching behaviour

wahjava:
At present, when a DNS record is cached, instead of caching records
present in the answer individually, current implementation caches the
answer and updates the TTL according to the lowest TTL present in the
records. e.g.

Actual reply from Google Public DNS resolver:

```sh
λ curl -s https://dns.google.com/resolve?name=something.blogspot.kr|jq .Answer
```

```json
[
  {
    "name": "something.blogspot.kr.",
    "type": 5,
    "TTL": 599,
    "data": "blogspot.l.googleusercontent.com."
  },
  {
    "name": "blogspot.l.googleusercontent.com.",
    "type": 1,
    "TTL": 299,
    "data": "172.217.166.225"
  }
]
```

Reply from dnscrypt-proxy using Google Public DNS as upstream:

```sh
λ drill something.blogspot.kr
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 34961
;; flags: qr rd ra ; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0
;; QUESTION SECTION:
;; something.blogspot.kr.       IN      A

;; ANSWER SECTION:
something.blogspot.kr.  298     IN      CNAME   blogspot.l.googleusercontent.com.
blogspot.l.googleusercontent.com.       298     IN      A       172.217.166.225

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 84 msec
;; EDNS: version 0; flags: ; udp: 512
;; SERVER: 127.0.0.2
;; WHEN: Sun Apr 14 06:35:29 2019
;; MSG SIZE  rcvd: 165
```

As you can see TTLs of all the records present in the response are
lowered to TTL of 300.

Following is the dnscrypt-proxy caching configuration used in this
test:

```toml
cache = true
cache_size = 512
cache_max_ttl = 86400
cache_min_ttl = 200
cache_neg_min_ttl = 60
cache_neg_max_ttl = 600
```

Apart from this current caching implementation doesn't cache other
records returned in the response when queried for them, i.e. in above
case only record for "something.blogspot.kr." is cached, and not
"blogspot.l.googleusercontent.com.", until
"blogspot.l.googleusercontent.com." is explicitly queried.

The commit here alters current caching implementation fixes both of
the above mentioned issues.

```sh
λ drill something.blogspot.kr.
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 37969
;; flags: qr rd ra ; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0
;; QUESTION SECTION:
;; something.blogspot.kr.       IN      A

;; ANSWER SECTION:
something.blogspot.kr.  573     IN      CNAME   blogspot.l.googleusercontent.com.
blogspot.l.googleusercontent.com.       252     IN      A       173.194.200.132

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 33 msec
;; EDNS: version 0; flags: ; udp: 4096
;; SERVER: 127.0.0.1
;; WHEN: Sun Apr 14 23:22:58 2019
;; MSG SIZE  rcvd: 165
λ drill something.blogspot.kr.
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 63006
;; flags: qr rd ra ; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0
;; QUESTION SECTION:
;; something.blogspot.kr.       IN      A

;; ANSWER SECTION:
something.blogspot.kr.  571     IN      CNAME   blogspot.l.googleusercontent.com.
blogspot.l.googleusercontent.com.       250     IN      A       173.194.200.132

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; EDNS: version 0; flags: ; udp: 4096
;; SERVER: 127.0.0.1
;; WHEN: Sun Apr 14 23:23:00 2019
;; MSG SIZE  rcvd: 112
λ drill blogspot.l.googleusercontent.com
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 4122
;; flags: qr rd ra ; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
;; QUESTION SECTION:
;; blogspot.l.googleusercontent.com.    IN      A

;; ANSWER SECTION:
blogspot.l.googleusercontent.com.       238     IN      A       173.194.200.132

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; EDNS: version 0; flags: ; udp: 4096
;; SERVER: 127.0.0.1
;; WHEN: Sun Apr 14 23:23:12 2019
;; MSG SIZE  rcvd: 77
```

### DIFF
--- a/dnscrypt-proxy/dnsutils.go
+++ b/dnscrypt-proxy/dnsutils.go
@@ -84,39 +84,56 @@ func StripTrailingDot(str string) string {
 	return str
 }
 
-func getMinTTL(msg *dns.Msg, minTTL uint32, maxTTL uint32, cacheNegMinTTL uint32, cacheNegMaxTTL uint32) time.Duration {
+func updateMsgTTLs(msg *dns.Msg, minTTL uint32, maxTTL uint32, cacheNegMinTTL uint32, cacheNegMaxTTL uint32) {
 	if (msg.Rcode != dns.RcodeSuccess && msg.Rcode != dns.RcodeNameError) || (len(msg.Answer) <= 0 && len(msg.Ns) <= 0) {
-		return time.Duration(cacheNegMinTTL) * time.Second
+		return
 	}
+
 	var ttl uint32
-	if msg.Rcode == dns.RcodeSuccess {
-		ttl = uint32(maxTTL)
-	} else {
-		ttl = uint32(cacheNegMaxTTL)
-	}
+
 	if len(msg.Answer) > 0 {
 		for _, rr := range msg.Answer {
+			if msg.Rcode == dns.RcodeSuccess {
+				ttl = uint32(maxTTL)
+			} else {
+				ttl = uint32(cacheNegMaxTTL)
+			}
 			if rr.Header().Ttl < ttl {
 				ttl = rr.Header().Ttl
 			}
+			if msg.Rcode == dns.RcodeSuccess {
+				if ttl < minTTL {
+					ttl = minTTL
+				}
+			} else {
+				if ttl < cacheNegMinTTL {
+					ttl = cacheNegMinTTL
+				}
+			}
+			rr.Header().Ttl = ttl
 		}
-	} else {
-		for _, rr := range msg.Ns {
-			if rr.Header().Ttl < ttl {
-				ttl = rr.Header().Ttl
+	}
+
+	for _, rr := range msg.Ns {
+		if msg.Rcode == dns.RcodeSuccess {
+			ttl = uint32(maxTTL)
+		} else {
+			ttl = uint32(cacheNegMaxTTL)
+		}
+		if rr.Header().Ttl < ttl {
+			ttl = rr.Header().Ttl
+		}
+		if msg.Rcode == dns.RcodeSuccess {
+			if ttl < minTTL {
+				ttl = minTTL
+			}
+		} else {
+			if ttl < cacheNegMinTTL {
+				ttl = cacheNegMinTTL
 			}
 		}
+		rr.Header().Ttl = ttl
 	}
-	if msg.Rcode == dns.RcodeSuccess {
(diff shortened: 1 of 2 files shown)


jedisct1:
The cache can certainly be improved, but per-record caching is very tricky to get right even in a non-recursive setup.

What happens when an `RRSIG` record expires before the record it represents? Should we invalidate the whole zone?

I'm also a little bit concerned about how this will play with server-side filters, as a record can be present both in a refused response, and in a normal one.

Recursive servers have to check that all the RRs are in-bailiwick before caching them in order to avoid injections. Here, we don't do anything similar, which doesn't make me comfortable.

From a compatibility standpoint, there are so many broken servers and load balancers, returning different responses according to the question, that I'd rather not cache records that don't directly match the question either.

Caching individual records also increase memory requirements.

On the implementation, there is no domain separation between hashes computed for individual records and for questions. That also may open a vulnerability.

It looks like `updateTTLs()` can set negative TTLs. Records from the additional section are not cached any more. Why not? Nothing prevents this section from having other record types than `OPT`.

In your example, when the `CNAME` expires, we have to perform a new query. When the `A` record the `CNAME` points to expires, a new query is needed as well. So, a query has to be made when `min(CNAME_TTL, A_TTL)` is reached no matter what.

Setting all the RRs to the minimum TTL, and caching per question may be suboptimal, but I feel like this is the safest option, from a reliability and security perspectives, as well as what maximizes compatibility with broken servers.

PR #820, opened 2019-05-13
Add support for building snaps

ghost:
I've opened this PR to request support for building a snap package of dnscrypt-proxy.

To give you a bit more context, snaps are cross-distro Linux software packages, supported on LTS and non-LTS Ubuntus, as well as Debian, Manjaro, Fedora, OpenSUSE, Arch, and many others distributions.

Snaps come with automatic updates, which could be quite useful for applications with focus on security and privacy, very much like dnscrypt-proxy. This means if you introduce an important update or a patch, it will propagate to your users, allowing you to maintain good control over the security element of the application (snap).

The Snap Store (snapcraft.io/store) has an audience of millions, so this could give extra coverage and visibility to your project.

Snaps can be built in several different ways: locally using the snapcraft cli tool, build/CI systems (e.g. Travis) or online via build.snapcraft.io, which is a free service for snap users - and it allows you to hook your GitHub repos directly and build for multiple architectures. You may be perticularly interested in ARM builds, if people are deploying dnscrypt-proxy to something like Raspberry Pi.

Here are the steps to get a snap built locally with snapcraft:

1. Install snapcraft tool and checkout the forked repo that has the snapcraft.yaml file for building the snap.

snap install snapcraft --classic --beta
git clone https://github.com/igorljubuncic/dnscrypt-proxy.git
cd dnscrypt-proxy
git checkout add-snapcraft
snapcraft

This command will generate a <dnscrypt-proxy-version>.snap file, something like dnscrypt-proxy_1.0_amd64.snap.

2. Test the snap.

dnscrypt-proxy can be installed and tested locally with:

snap install dnscrypt-proxy_1.0_amd64.snap --dangerous

The --dangerous flag is necessary because the snap isn't in the store just yet, and therefore isn't signed.

Once installed, the dnscrypt-proxy command can be executed, e.g.: snap run dnscrypt-proxy <options>

3. Register snap name and upload snap to the store.

- Register a developer account in the snap store https://snapcraft.io/account.
- Register the dnscrypt-proxy name in the store. 

snapcraft login
snapcraft register

- Upload this snap to the store. We use channels to differentiate risk levels - edge, beta, candidate, stable. You can promote to other channels once you're happy with the tests.

snapcraft push dnscrypt-proxy_1.0_amd64.snap --release edge

4. Test the version available in the store

You can do this on a clean machine (install from store and run). No need for the --dangerous flag anymore.

snap install dnscrypt-proxy --edge

5. Interfaces

The one last thing to note is, I built and tested dnscrypt-proxy with several interfaces defines - snaps are confined but can talk to the system through the mechanism of interfaces (https://docs.snapcraft.io/interface-management). For this one, we have:

- network (ability to access network)
- network-bind (ability to bind to ports)
- network-control (ability to control network devices)

The last one is not auto-connected and requires a manual connection on the command line (via snap connect), for security reasons, but once you open the account and register the snap, we can work on adding an assertion that will allow this to happen automatically when the dnscrypt-proxy snap is installed.

Once (and if) you are happy, you can push a stable release to the stable channel, update the store page, and promote the application online. We can help there, and we'd be glad to feature your application in our store. Thank you!

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,25 @@
+name: dnscrypt-proxy
+version: '1.0'
+summary: A flexible DNS proxy, with support for encrypted DNS protocols
+description: |
+  A flexible DNS proxy, with support for encrypted DNS protocols.
+grade: stable
+confinement: strict
+base: core18
+parts:
+  dnscrypt-proxy:
+    plugin: go
+    source: https://github.com/jedisct1/dnscrypt-proxy.git
+    go-importpath: github.com/jedisct1/dnscrypt-proxy
+    build-packages:
+      - build-essential
+apps:
+  dnscrypt-proxy:
+    command: bin/dnscrypt-proxy
+    daemon: simple
+    restart-condition: on-abnormal
+    plugs:
+      - home
+      - network
+      - network-bind
+      - network-control


jedisct1:
Hi,

And thanks for providing this!

Once installed, the snap doesn't seem to be usable. Is there a way to install the example configuration files? Where are the actual configuration files supposed to be located?

I manually downloaded a copy of the example configuration file, renamed it, and tried to run the snap. It originally didn't work because the configuration file apparently needs to belong to `root`.

Once done, the server list caches cannot be accessed:

```
[2019-05-22 16:14:28] [WARNING] /home/ubuntu/src/dnscrypt-proxy/dnscrypt-proxy/public-resolvers.md: open sf-27wkcgzg63i3lv5r.tmp: permission denied
[2019-05-22 16:14:28] [WARNING] /home/ubuntu/src/dnscrypt-proxy/dnscrypt-proxy/public-resolvers.md.minisig: open sf-oycxldg2oqzpu4pi.tmp: permission denied
```

I couldn't install the service either:

```
[2019-05-22 16:14:28] [FATAL] Failed to install DNSCrypt client proxy: open /etc/systemd/system/dnscrypt-proxy.service: permission denied
```

How to make this work?

ghost:
Hello,

Can you share the sequence of commands you ran?

In the snapcraft.yaml file, I specified this:

```apps:
  dnscrypt-proxy:
    command: bin/dnscrypt-proxy
    daemon: simple
    restart-condition: on-abnormal
    plugs:
      - home
      - network
      - network-bind
      - network-control
```

And this means there is a simple service that should run. However, if you need to bind privileged ports, the snap needs to run as sudo or root. For network-control, this will not be automatically connected for security reason (once the snap is in the store, you can request a change).

```snap interfaces dnscrypt-proxy
Slot           Plug
:home          dnscrypt-proxy
:network       dnscrypt-proxy
:network-bind  dnscrypt-proxy
-              dnscrypt-proxy:network-control
```

So you can do snap connect dnscrypt-proxy:network-control.

If you want to add files (to be included with the snap), you can use the dump plugin:
https://docs.snapcraft.io/dump-plugin

So you can dump the config file and move it under etc (relative to the snap).

You can inspect the contents of the snap you built by unsquashing the filesystem:

```
unsquashfs "snap name"
```

Or during the build process run `snapcraft --shell-after`, so you step into the build instance and see exactly what was created and how.

Can you try and see if this helps.

jedisct1:
The first thing I tried is `sudo snap run dnscrypt-proxy`, but since there was no configuration file, I downloaded the example one:

```sh
curl -L https://raw.githubusercontent.com/jedisct1/dnscrypt-proxy/master/dnscrypt-proxy/example-dnscrypt-proxy.toml > dnscrypt-proxy.toml
```

And ran

```sh
sudo snap run dnscrypt-proxy
```

That failed with:

```text
[2019-05-23 11:53:51] [FATAL] open /home/ubuntu/tmp/dnscrypt-proxy.toml: permission denied
```

Changing ownership on that file fixed it:

```sh
sudo chown 0:0 dnscrypt-proxy.toml
sudo snap run dnscrypt-proxy
```

However, cache files could not be written:

```text
[2019-05-23 11:57:07] [WARNING] /home/ubuntu/tmp/public-resolvers.md: open sf-eyf5poj27mvfgvek.tmp: permission denied
[2019-05-23 11:57:07] [WARNING] /home/ubuntu/tmp/public-resolvers.md.minisig: open sf-iqxoxcetm3prrlzm.tmp: permission denied
```

Changing ownership of the directory fixed it:

```sh
chown -R 0:0 .
sudo snap run dnscrypt-proxy
```

```text
[2019-05-23 11:57:49] [NOTICE] Source [public-resolvers.md] loaded
[2019-05-23 11:57:49] [NOTICE] dnscrypt-proxy 2.0.23
[2019-05-23 11:57:49] [NOTICE] Now listening to 127.0.0.1:53 [UDP]
[2019-05-23 11:57:49] [NOTICE] Now listening to 127.0.0.1:53 [TCP]
[2019-05-23 11:57:49] [NOTICE] Now listening to [::1]:53 [UDP]
[2019-05-23 11:57:49] [NOTICE] Now listening to [::1]:53 [TCP]
...
```

But I could not install the service:

```sh
sudo snap run dnscrypt-proxy -service install
```

```text
[2019-05-23 11:58:43] [NOTICE] Source [public-resolvers.md] loaded
[2019-05-23 11:58:43] [NOTICE] dnscrypt-proxy 2.0.23
[2019-05-23 11:58:43] [FATAL] Failed to install DNSCrypt client proxy: open /etc/systemd/system/dnscrypt-proxy.service: permission denied
```

jedisct1:
With `confinement: classic`, installing/stopping/restarting the service works, as well as the `user_name=` property to run the command as a non-root user.

ghost:
Classic gives you the same level of permission as the host, but typically, classic confinement is only used when there's no alternative.

For the service install, I would suggest the following:

Use the `dump` plugin:

Dump the configuration file.
Dump the dnscrypt-proxy.service file.
Use the `organize` keyword to move these two to the correct places.

For instance:

```
  myfile:
    plugin: dump
    source: .
    organize:
      file-to-move: etc/file-to-move
    stage: [ etc/file-to-move ]
```
This way, you won't need to "install" the service.

Can you test this?

jedisct1:
There is no static `dnscrypt-proxy.service` file. It's dynamically created by an external Go module, and includes the directory where the `-service iinstall` command was run from, so that config files can be found later.

Installing a DNS proxy on Linux is already very complicated. I'd rather avoid having completely different instructions according to how the software was installed.

ghost:
OK ... I was just thinking of having a portable setup (and confined). If you use classic, if things break on the host, you don't benefit from the advantages of snap isolation and updates. If you don't mind, can you give me a complex scenario of how you'd see dnscrypt-proxy deployed, and I'll test to see what's the best way to get it organized.

jedisct1:
Most people using this on Linux are not Linux experts, and I don't use Linux much either.

So, things should remain simple and consistent with the same software used in other environments.

If a configuration file is present, but the app says that it can't load it, this is confusing.

If the service cannot be installed in spite of having used `sudo`, this is frustrating.

If switching to a non-root user doesn't work even though the option has been enabled (which is what happens without `classic`), it looks like the app is quite buggy.

If caches and log files don't work even though the configuration files look fine, and work on different environment, the package will be immediately and definitely uninstalled.

Snaps can be interesting, especially if they can safely and automatically auto-updated. But using them shouldn't be a completely different experience than the regular packages.

ghost:
I agree - usage should be simple and consistent. I did not imply you should break the dev/work flow. I just want to see if there's a way to make things run smoothly under confinement (to which end I asked if you have a complex example I could use). If things don't run smoothly, there's always the option to use classic confinement (for use cases where strict breaks things).

PR #828, opened 2019-05-21
make generate-domains-blacklist.py compatible to both python2 and python3

simonfxr:
Pretty much as the title says. I also fixed a dead URL in domains-blacklist-all.conf, it moved to Gitlab [LINK](https://github.com/quidsup/notrack/)


jedisct1:
Python2 is [almost dead already](https://pythonclock.org/).

All the major projects [stopped supporting python 2](https://python3statement.org/) to encourage the switch to Python 3 before its end of life, scheduled to happen in a couple months.

I don't think we should pedal backwards.

simonfxr:
@jedisct1 The reason for this PR is that it previously didn't work for me under python3. AFAIK urllib2 is python2 only, so previously it wouldn't run under python3 but now it does run under python2 and python3.

jedisct1:
Thank you!

PR #829, opened 2019-05-22
Whitelist Server Source Domains

EncryptTown:
A malicious blacklist, or accidental block, could prevent dnscrypt-proxy users from being able to fetch public resolvers and important certificate updates. Both URLs are taken from the default config:

[sources.'public-resolvers']
urls = ['https://raw.githubusercontent.com/DNSCrypt/dnscrypt-resolvers/master/v2/public-resolvers.md', 'https://download.dnscrypt.info/resolvers-list/v2/public-resolvers.md']


jedisct1:
Thank you!

PR #833, opened 2019-05-26
Fix synthentic (incl. forwarded) responses not getting processed by repsonse plugins

fholzer:
In case one of the request plugins creates a synthetic response, (e.g. forwarding plugin)
the response plugins wouldn't be called at all. This means that
such responses would not be processed by the nxlog, block_ip(v6)
and cache response plugins. In the case of block plugins the response simply wouldn't be blocked. Due to the cache response plugin not being called none of the responses generated by the forwarding plugin are cached.

Note: with this PR as it is now, the cache response plugin will also be called for responses generated by the cache plugin, which I suppose isn't desirable.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -351,7 +351,6 @@ func (proxy *Proxy) processIncomingQuery(serverInfo *ServerInfo, clientProto str
 		pluginsState.returnCode = PluginsReturnCodeForward
 	}
 	if len(response) == 0 && serverInfo != nil {
-		var ttl *uint32
 		if serverInfo.Proto == stamps.StampProtoTypeDNSCrypt {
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi Ferdinand,

This is the expected behavior. A synthetic response is something that should be delivered to the client.

These shouldn't be cached, as a synthetic response can depend on properties that are not included in the cache key.

Forwarding is typically done to access local, trusted domains, so I'm not sure that applying blocking rules is desirable.

Maybe we need to introduce new states, to differentiate synthetic responses from forwarded/cloaked responses. But I'd rather keep synthetic responses as final responses that bypass other filters.

fholzer:
>Forwarding is typically done to access local, trusted domains, so I'm not sure that applying blocking rules is desirable.

Sure, didn't think of that. Personally, I'm using it also for some non-local domains, so it made sense to me.

Cheers/f

PR #834, opened 2019-05-26
Enhance logging

fholzer:
Enhance query logging by adding:
* Total request time
   The duration between dnscrypt-proxy receiving the request and it sending the response
* Upstream request time
   The duration between sending a request to an upstream DNS server and us receiving the response. Including requests forwarded based on `forwarding_rules`.

### DIFF
--- a/dnscrypt-proxy/plugin_forward.go
+++ b/dnscrypt-proxy/plugin_forward.go
@@ -5,6 +5,7 @@ import (
 	"math/rand"
 	"net"
 	"strings"
+	"time"
 	"unicode"
 
 	"github.com/jedisct1/dlog"
@@ -92,7 +93,9 @@ func (plugin *PluginForward) Eval(pluginsState *PluginsState, msg *dns.Msg) erro
 		return nil
 	}
 	server := servers[rand.Intn(len(servers))]
+	reqStart := time.Now()
 	respMsg, err := dns.Exchange(msg, server)
+	pluginsState.forwardDuration = time.Now().Sub(reqStart)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


jedisct1:
The duration is difficult to parse due to variable units. Can we get them all as a quantity of the same unit?

fholzer:
Sure. Wasn't sure, if you want it human readable in the tsv, and machine readable in ltsv...
Is nanoseconds fine, or overkill. (That's what i used for ltsv)

jedisct1:
Milliseconds are probably granular enough :)

jedisct1:
Thanks!

Do we need both values?

I've been running this a bit, with and without forwarding, and the numbers are always really close.

When would they substantially differ?

fholzer:
>I've been running this a bit, with and without forwarding, and the numbers are always really close.
>When would they substantially differ?
Point taken. When using a unit smaller than milliseconds you'd see a difference. Yes they should be really close all the time. If they wouldn't be, then that would point to some issue with dnscrypt-proxy I suppose. So it might come in handy for debugging issues, but not for regular use, true.

You want me to remove one of the two? If so, which one?

On another note, I was thinking about adding an additional field to the query log that reflects whether the response was served from cache or not. Would you be interested in that? (Maybe only in ltsv?)
Something like `cachehit:true`/`cachehit:false`.

fholzer:
Also: thank you for maintaining this project!

jedisct1:
Maybe keep only the total time, especially since the code path is simpler?

Having a `cachehit` property (maybe just `hit`?) in the LTSV file can indeed be interesting.

PR #835, opened 2019-05-26
Add support for removing EDNS client subnet information

fholzer:
Also add logging of EDNS client subnet information to query log.

### DIFF
--- a/dnscrypt-proxy/plugin_edns.go
+++ b/dnscrypt-proxy/plugin_edns.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"github.com/miekg/dns"
+)
+
+type PluginEdns struct {
+	removeEdnsClientSubnet bool
+}
+
+func (plugin *PluginEdns) Name() string {
+	return "edns"
+}
+
+func (plugin *PluginEdns) Description() string {
+	return "Remove EDNS client subnet data from requests"
+}
+
+func (plugin *PluginEdns) Init(proxy *Proxy) error {
+	plugin.removeEdnsClientSubnet = proxy.removeEdnsClientSubnet
+	return nil
+}
+
+func (plugin *PluginEdns) Drop() error {
+	return nil
+}
+
+func (plugin *PluginEdns) Reload() error {
+	return nil
+}
+
+func (plugin *PluginEdns) Eval(pluginsState *PluginsState, msg *dns.Msg) error {
+	edns := msg.IsEdns0()
+	if edns == nil {
+		return nil
+	}
+
+	// extract client subnet details for logging
+	for _, o := range edns.Option {
+		switch o.(type) {
+		case *dns.EDNS0_SUBNET:
+			subnet := o.(*dns.EDNS0_SUBNET)
+
+			switch subnet.Family {
+			case 1: // IPv4
+				fallthrough
+			case 2: // IPv6
+				pluginsState.ednsClientIP = subnet.Address
+				pluginsState.ednsClientMask = subnet.SourceNetmask
+			}
+		}
+	}
+
+	// remove client subnet details from request
+	if plugin.removeEdnsClientSubnet {
+		plugin.removeClientSubnet(edns)
+	}
+
+	return nil
+}
+
+func (plugin *PluginEdns) removeClientSubnet(edns *dns.OPT) {
+	i := 0
+	for _, o := range edns.Option {
+		switch o.(type) {
+		case *dns.EDNS0_SUBNET:
+		default:
+			edns.Option[i] = o
+			i++
+		}
+	}
+	edns.Option = edns.Option[:i]
+}
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -75,6 +75,7 @@ type Config struct {
 	OfflineMode              bool                       `toml:"offline_mode"`
 	HTTPProxyURL             string                     `toml:"http_proxy"`
 	RefusedCodeInResponses   bool                       `toml:"refused_code_in_responses"`
+	RemoveEdnsClientSubnet   bool                       `toml:"remove_edns_client_subnet"`
 }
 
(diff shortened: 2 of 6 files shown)


jedisct1:
Interesting. What is your use case for this?

ECS is typically added by recursive resolvers. For example Open DNS will send a copy of your network address to other companies. The fact that the client query included ECS or not is completely ignored.

Where does that ECS information come from in your case?

fholzer:
In my home network I'm using dnsmasq on my Linksys router. It's running OpenLEDE (an OpenWRT fork). I'd love to keep it that way because it'll take care of dynamic DNS records for any connected devices and setting the appropriate search domain on them. This means I can't tell clients to directly use dnscrypt-proxy. So because all the requests are first received by dnsmasq, then forwarded to dnyscrypt-proxy I don't see the actual client device IP, even when enabling EDNS client subnet=32 on dnsmasq, because dnscrypt-proxy currently isn't evaluating it.

This PR would do exactly that. Evaluate the client subnet, and also provide an option to remove it. Upstream DNS servers don't need to know about my internal IPs ;)

fholzer:
I guess I missed one important thing: ultimately the reason is that I'd like to ship the query log to ELK so I can visualize it. And in case there's something fishy, I'd like to know which client on my network sent that request. With my network setup, see previous message, that's only possible with this patch.

That being said, I suppose it's not a common setup.
If you're still interested let me know, I would rebase against master.

jedisct1:
I see. I use OpenLEDE as well, but the other way round, with dnscrypt-proxy as a front-end, forwarding local zones to dnsmasq.

`removeEdnsClientSubnet` is a bit confusing, as this is mainly for logging. Maybe we can find a better name?

In the TSV output, would it make sense to replace the `host` part with ECS information instead of having another field?

The TSV output is designed to be human-readable, and is displayed on small screens (such as in DNSCloak in iOS), so we shouldn't put too much information here, especially information that is only available in very specific use cases.

The LTSV output, on the other hand, can be extended at will, and fields that are not used can be omitted.

dnscrypt-proxy/plugin_edns.go
```diff
@@ -0,0 +1,73 @@
...
+	i := 0
+	for _, o := range edns.Option {
+		switch o.(type) {
+		case *dns.EDNS0_SUBNET:
+		default:
+			edns.Option[i] = o
+			i++
+		}
+	}
+	edns.Option = edns.Option[:i]
```
jedisct1, line 72: The query can include multiple EDNS options. We should only remove the ECS ones, and keep the other ones.

fholzer: Right. So that's why there's the case statement that filters for options of type EDNS0_SUBNET. As golang doesn't fall through if not explicitly instructed to it won't enter the default block for SUBNET options.
For any other EDNS option it will go directly to the default block, which basically tells it to not remove this option.
This is nifty way to remove specific elements of a slice while iterating over it I found at https://stackoverflow.com/a/20551116/320944

fholzer:
>removeEdnsClientSubnet is a bit confusing, as this is mainly for logging. Maybe we can find a better name?

Right, so I thought about that myself. There's actually two parts to this.
1. Extract EDNS subnet option for writing it to the query log.
   You'd want to do this as early in the request plugin chain as possible, so that whatever gets logged already has that information in it.
2. Removing EDNS subnet option before sending this request anywhere else.
   You'd want to do this after the cached request plugin, so that you don't rewrite the EDNS options slice for all requests even if they would be answered from the cache.

So maybe it would make sense to split this into two plugins. The first, extracting the information always enabled. The second, removing the information only added to the request plugin slice if `removeEdnsClientSubnet` is enabled via config file...

What do you think?

privacyguy123:
https://www.quad9.net/faq/#What_is_EDNS_Client-Subnet

MystesofEternity:
Was this completely forgotten?

ghost:
Hello.

I just want to write, that one of the Provider mentions about `ECS` and suggest, that if User wants to he "_(...) may opt out this behavior if you use doh-client by setting no_ecs = true_" (please see "2. Information we share" [in:] https://dns.sb/privacy/).

Generally, `ECS` seems to be related with DoH (what about DNSCrypt?).  

Best regards.

ReSearchITEng:
my 2 cent: this [doh-proxy](https://github.com/tinkernels/doh-proxy) project has something like: "-edns-subnet"

hugepants:
>So because all the requests are first received by dnsmasq, then forwarded to dnscrypt-proxy I don't see the actual client device IP

@fholzer If you're using dnsmasq at the front-end which forwards external DNS queries to dnscrypt-proxy (as documented in the [Wiki](https://github.com/DNSCrypt/dnscrypt-proxy/wiki/Installation-on-OpenWrt)), then why not simply enable query logging for dnsmasq, which will log the original requesting client IP, its DNS query and the upstream response via dnscrypt-proxy (when applicable)?

in OpenWrt/LEDE's `/etc/config/dhcp`:
```sh
config dnsmasq
    option logqueries '1'    # equivalent to --log-queries=extra in dnsmasq
    option logfacility '/tmp/dnsmasq_queries.log'    # very chatty, best to write to an external share
```

hugepants:
In situations where `dnscrypt-proxy` is not used at the front-end e.g. with `dnsmasq`, `Unbound` and similar; `dnscrypt-proxy` cannot log the requesting client IP address. The simplest and cleanest solution is to enable verbose query logging in the front-end application which will capture the client IP as well as the DNS query details. I have added a [section in the Wiki](https://github.com/DNSCrypt/dnscrypt-proxy/wiki/Installation-on-OpenWrt#logging-dns-queries-with-client-ips) explaining how to do this in OpenWrt with `dnsmasq`.

Using EDNS client subnet information to accomplish this is overkill and adds unnecessary complexity to `dnscrypt-proxy`, so I'm closing this.

PR #838, opened 2019-05-28
Further log enhancements

fholzer:
* Remove request forwarding measurement from log
* Log whether response was served from cache (only in ltsv)


jedisct1:
Fantastic, thank you!

PR #839, opened 2019-05-29
regex isn't fully parsing the complicated csv file

EncryptTown:
Not all URLs are extracted from the complicated csv file.
However, they do offer a txt file for the same list, which does work correctly with the current regex:
https://www.malwaredomainlist.com/forums/index.php?topic=3270.0
This url replacement pull request is easier than rewriting the entire regex (which then breaks other lists).


jedisct1:
Thank you!

PR #861, opened 2019-06-18
Initial support for DNS64 (RFC 6147)

s-s:
Initial support for synthesizing AAAA records from A records for IPv6-only networks (i.e. Rogers in Canada).
Can be configured by a set of static prefixes or ISP's DNS64 resolvers for prefix discovery purposes (RFC 7050).


jedisct1:
Oh, this is fantastic, thanks!

Since you mention that this is only initial support, what is missing? Should it be merged now?

s-s:
Well, as of now there is no DNSSEC processing, no PTR handling and CNAME following from RFC 6147.
Also, as for RFC 7050 implementation: prefixes discovery is executed only once during plugin initialization (not periodically by following TTL) and there are no prefixes and connectivity validations.

jedisct1:
Thanks. Implementing everything is indeed quite a bit of work, but what you did is the most important part and is already useful. Following `CNAME` seems quite of a requirement, but this is going to be trickier to implement.

DNSSEC will obviously be broken by this.

jedisct1:
Should I merge it now?

unixfox:
Is there any plan to merge this PR? I would love to stop using bind9 as a DNS64 server.

PR #864, opened 2019-06-24
Hacky support for socks5 proxying FetchCurrentDNSCryptCert

### LINKED ISSUE #817: FetchCurrentDNSCryptCert does not honor proxy settings
Currently [FetchCurrentDNSCryptCert](https://github.com/jedisct1/dnscrypt-proxy/blob/0a535e28ab9c39a7123720419f35ee24f796bb4b/dnscrypt-proxy/dnscrypt_certs.go#L23) does not honor proxy settings causing it to try direct connection regardless of these settings.

It seems that miekg/dns does not currently support proxy dialers (Unless I missed that), so some changes would be required to support that as well.

z0w13:
This requires a [patched miekg/dns](https://github.com/z0w13/dns/tree/socks5) as well and would resolve #817 if implemented cleanly.

I currently don't have the resources / know-how to cleanly implement this so I'm providing it as a starting point and/or hack for people that want this functionality


jedisct1:
Nice!

Do you think you could get the changes to `miekg/dns` merged upstream?

z0w13:
Unless it's cleaned up and doesn't break existing functionality, 2 issues right now are 

1. Not supporting TLS dialing ([see here](https://github.com/z0w13/dns/blob/362ce57eff6fb8f52768f9bbb7b1c90f980dabe4/client.go#L104))
2. Not setting the requestTimeout from dialer timeout ([here](https://github.com/z0w13/dns/commit/d0d9595af7f3cdcbd99e12eb8562d57dacf0d2dd#diff-d0b7f86cea13f90d3cc0cc2338da8d5fL323))

jedisct1:
I think this is supported now. Through SOCKS, Tor and Relays.

PR #868, opened 2019-06-27
Option to redirect refused responses to a given IP address

jrnewell:
This might be too specific of a use case, so I won't be offended if you don't want to merge this.

I run a pixelserv-tls server (https://github.com/kvic-z/pixelserv-tls) to respond to add/trackers with a blank pixel, so I modified the option `refused_code_in_responses` to be able to take an IP address (in addition to 'refused' or the 'hinfo' response similar to before).

Since the config option is a new type (string vs bool), I changed the name to `blocked_query_response` but still support `refused_code_in_responses` with a deprecated message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 *~
 dnscrypt-proxy/dnscrypt-proxy2
 dnscrypt-proxy/dnscrypt-proxy
+.idea
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -81,6 +81,7 @@ type Config struct {
 	OfflineMode              bool                       `toml:"offline_mode"`
 	HTTPProxyURL             string                     `toml:"http_proxy"`
 	RefusedCodeInResponses   bool                       `toml:"refused_code_in_responses"`
+	RespondWithIP            string                     `toml:"respond_with_ip"`
 }
 
 func newConfig() Config {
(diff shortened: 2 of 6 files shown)


jedisct1:
This is a good idea!

To be consistent with other options, maybe `blocked_query_response` shouldn't directly take an IP address, but something like `A:<ip address>` or `AAAA:<ip address>`.

Speaking of which, maybe the ability to specify both an IPv4 address and an IPv6 address is something people will want. With the IPv4 address being mandatory, and the IPv6 one defaulting to the 4-in-6 address if not provided.

jrnewell:
Cool! I can make those changes.  So, are you envisioning that it would be an array of strings type then, so valid options would be `blocked_query_response = [ 'refused' ]` or `blocked_query_response = [ 'hinfo' ]` or `blocked_query_response = [ 'A:192.168.1.4' ]` or `blocked_query_response = [ 'A:192.168.1.4', 'AAAA:2001:0db8:85a3:0000:0000:8a2e:0370:7334' ]`

jedisct1:
An array doesn't feel very natural in most cases. So maybe something like:

```toml
blocked_query_response = 'refused'
blocked_query_response = 'hinfo'
blocked_query_response = 'a:192.168.1.4,aaaa:[2001:0db8:85a3:0000:0000:8a2e:0370:7334]'
```
what do you think?

jrnewell:
yeah, that sounds good.  I went ahead and made those changes

jedisct1:
Awesome!

Thank you so much for that great feature!

PR #877, opened 2019-07-09
Use ReadMsgUDP/WriteMsgUDP instead

maskedeken:
It is compatible for iptables REDIRECT/DNAT on Linux


jedisct1:
That code is a copy&paste from `miekg/dns` (that is likely to break on Windows, btw).

If this is something all UDP-based servers need, it might be more interesting to maintain a clean, standalone, portable module that provides this functionality.

PR #891, opened 2019-07-22
Cloaking example: DuckDuckGo safe search

javabean:
As documented at https://help.duckduckgo.com/features/safe-search/


jedisct1:
Thank you! I didn't know they also supported this. Great!

tunip:
There are some other subfomains for duckduckgo.com like proxy.duckduckgo.com, icons.duckduckgo.com, help.duckduckgo.com which will not work with this example cloak

I suggest this for a working cloaking:

```
=duckduckgo.com           safe.duckduckgo.com
```

jedisct1:
Oh, good catch, thank you @tunip

Issue #949, opened 2019-10-06
Proposal: allow continuing if at least one resolver is okay


alisonatwork:
This morning I encountered a problem where `dnscrypt-proxy` did not start up due to a single bad entry in the public resolvers file.

```
PS C:\ProgramData\chocolatey\lib\dnscrypt-proxy\tools\win64> .\dnscrypt-proxy.exe
[2019-10-06 10:13:25] [NOTICE] dnscrypt-proxy 2.0.27
[2019-10-06 10:13:25] [NOTICE] Network connectivity detected
[2019-10-06 10:13:25] [NOTICE] Source [public-resolvers.md] loaded
[2019-10-06 10:13:25] [ERROR] Invalid or unsupported stamp: [sdns://AgMAAAAAAAAAJzIwMDE6MTlmMDo3MDAxOjU1NTQ6NTQwMDowMmZmOmZlNTc6MzA3NyBsA2QQ3lR1Nl9Ygfr8FdBIpL-doxmHECRx3T5NIXYYtxNkbnMuY29udGFpbmVycGkuY29tCi9kbnMtcXVlcnk]
[2019-10-06 10:13:25] [CRITICAL] Unable to use source [public-resolvers]: [Invalid stamp (IP address)]
[2019-10-06 10:13:25] [FATAL] Invalid stamp (IP address)
PS C:\ProgramData\chocolatey\lib\dnscrypt-proxy\tools\win64> notepad .\public-resolvers.md
PS C:\ProgramData\chocolatey\lib\dnscrypt-proxy\tools\win64> .\dnscrypt-proxy.exe
[2019-10-06 10:14:26] [NOTICE] dnscrypt-proxy 2.0.27
[2019-10-06 10:14:26] [NOTICE] Network connectivity detected
[2019-10-06 10:14:26] [NOTICE] System DNS configuration not usable yet, exceptionally resolving [raw.githubusercontent.com] using fallback resolver [114.114.114.114:53]
[2019-10-06 10:14:27] [NOTICE] Source [public-resolvers.md] loaded
...
```

The error is shown as (`Invalid or unsupported stamp`). I resolved it by commenting out the bugged entry and restarting. I don't think the comment made a difference, but by changing the signature it forced a re-download of the public resolvers file.

[...]

PR #950, opened 2019-10-06
when parsing a source, try to process all of the servers to ensure maximum connectivity

alisonatwork:
This is a follow-up to #949 - now that a partial failure in the source does not cancel startup, it would be good to process as many servers as possible, to avoid the case where there a parse error in a server early on the list and then we end up thrashing all the early servers.

This PR still aborts parsing immediately if the source file is badly corrupt (i.e. no name/description).

However, for the case where there is a duplicate, missing or invalid server stamp, we log at warn level and collect the error messages to return as a single error at the end.

Note I have removed the part of the log message `in source from [%v]`, since that created a noisy log and I think the log in `config.go` around line 590 will provide more useful information. Can add it back if required.

Here is an example of the output with a messed up file:

```
$ ./dnscrypt-proxy.exe
[2019-10-06 18:26:57] [NOTICE] dnscrypt-proxy 2.0.27
[2019-10-06 18:26:57] [NOTICE] Network connectivity detected
[2019-10-06 18:26:57] [NOTICE] Source [public-resolvers.md] loaded
[2019-10-06 18:26:57] [WARNING] Invalid or unsupported stamp [sdns://AgMAAAAAAAADjE2OC4yMzUuODEuMTY3ID4aGg9sU_PpekktVwhLW5gHBZ7gV6sVBYdv2D_aPbg4EmRucy1ueWMuYWFmbGFsby5tZQovZG5zLXF1ZXJ5]: illegal base64 data at input byte 114
[2019-10-06 18:26:57] [WARNING] Multiple stamps for server [adguard-dns-doh]
[2019-10-06 18:26:57] [WARNING] Missing stamp for server [adguard-dns-family-ipv6]
[2019-10-06 18:26:57] [WARNING] Error in source [public-resolvers]: [Invalid or unsupported stamp [sdns://AgMAAAAAAAADjE2OC4yMzUuODEuMTY3ID4aGg9sU_PpekktVwhLW5gHBZ7gV6sVBYdv2D_aPbg4EmRucy1ueWMuYWFmbGFsby5tZQovZG5zLXF1ZXJ5]: illegal base64 data at input byte 114, Multiple stamps for server [adguard-dns-doh], Missing stamp for server [adguard-dns-family-ipv6]] -- Continuing with reduced server count [204]
...
```

### DIFF
--- a/dnscrypt-proxy/sources.go
+++ b/dnscrypt-proxy/sources.go
@@ -217,12 +217,19 @@ func (source *Source) Parse(prefix string) ([]RegisteredServer, error) {
 
 func (source *Source) parseV2(prefix string) ([]RegisteredServer, error) {
 	var registeredServers []RegisteredServer
+	var stampErrs []string
+	appendStampErr := func(format string, a ...interface{}) {
+		stampErr := fmt.Sprintf(format, a...)
+		stampErrs = append(stampErrs, stampErr)
+		dlog.Warn(stampErr)
+	}
 	in := string(source.in)
 	parts := strings.Split(in, "## ")
 	if len(parts) < 2 {
 		return registeredServers, fmt.Errorf("Invalid format for source at [%v]", source.urls)
 	}
 	parts = parts[1:]
+PartsLoop:
 	for _, part := range parts {
 		part = strings.TrimFunc(part, unicode.IsSpace)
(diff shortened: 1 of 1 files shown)


jedisct1:
Thank you!

PR #969, opened 2019-10-18
Cleanups and small fixes

maage:
I guess most of these fixes can be picked individally too.

Most are just cleanups where I try to modify code to match my view of how things should be.

Fixes are bugfixes.

Changes modify current functionality and can break something. Or give error where there were none before.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -13,6 +13,7 @@ import (
 	"github.com/jedisct1/dlog"
 	clocksmith "github.com/jedisct1/go-clocksmith"
 	stamps "github.com/jedisct1/go-dnsstamps"
+	"github.com/miekg/dns"
 	"golang.org/x/crypto/curve25519"
 )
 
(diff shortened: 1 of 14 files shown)


jedisct1:
Wow, this is amazing!

Thanks a ton for having taken the time to review that code and add all these improvements!

This is my first project in Go, so these changes are very helpful.

Thanks again!

dnscrypt-proxy/proxy.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/jedisct1/dlog"
 	clocksmith "github.com/jedisct1/go-clocksmith"
 	stamps "github.com/jedisct1/go-dnsstamps"
+	"github.com/miekg/dns"
```
dapphp, line 16: I know nothing about Go, but am wondering if pulling in a fairly large library for one rcode is worth it or has any negative side effects (increase binary size).

jedisct1: It's already used for many more things in other files :)

jedisct1:
Honestly one of the best PR I've ever received.

PR #970, opened 2019-10-18
feature: service_linux: Support systemd watchdog

maage:
Can be easily tested with this kind of command:

```
systemctl reset-failed --user
systemd-run --user --unit=foo --property={NotifyAccess=main,WatchdogSec=1m} --working-directory="$(readlink -f dnscrypt-proxy)" "$(readlink -f dnscrypt-proxy)"/dnscrypt-proxy-linux-amd64 -config example-dnscrypt-proxy.toml
```

Without this patch program is killed after minute. Like this:

Oct 18 22:15:43 workstation.lan systemd[1218]: foo.service: Watchdog timeout (limit 1min)!

I have this running with this in dnscrypt-proxy.service:

```
[Service]
NotifyAccess=main
WatchdogSec=3m
Restart=on-failure
...
```

As I mention in TODO comment, this probably does not hande every problem but I guess it does not hurt. If there is other internal metric to be sure the program is working it should transmitted there. As there is no such thing now, I used this simplest setup.

I decided to signal watchdog 3 times per period, just to be sure.

### DIFF
--- a/dnscrypt-proxy/privilege_linux.go
+++ b/dnscrypt-proxy/privilege_linux.go
@@ -47,7 +47,9 @@ func (proxy *Proxy) dropPrivilege(userStr string, fds []*os.File) {
 		dlog.Fatal(err)
 	}
 
-	ServiceManagerReadyNotify()
+	if err := ServiceManagerReadyNotify(); err != nil {
+		dlog.Fatal(err)
+	}
 
 	args = append(args, "-child")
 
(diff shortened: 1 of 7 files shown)


maage:
Ah, fixed and pushed again to take into account android.

jedisct1:
Neat!

But instead of reimplementing this everywhere, wouldn't it make sense to have it in `kardianos/service` instead?

maage:
kardianos/service does not currently have coreos/go-systemd/daemon as dependency. It's systemd part is mainly managin systemd services using systemctl command. So as far as I can see it is not good fit here.

My patch just uses already used coreos/go-systemd just slightly further.

Issue #924, opened 2019-09-07
How to debug SERVER_ERROR?


PeterDaveHello:
Sometimes there are many `SERVER_ERROR` in the query log, but I can't see any related issues, I see no options in the config and parameters to help debug, even `log_level` is set to `0` already, I know reading the source code might be one method but it'll be great if there can be some more friendly ways to do it. Many thanks.

jedisct1:
`SERVER_ERROR` is usually a timeout. A query was sent to the server, but no reply was received.

AnyWAT:
[2019-09-08 06:28:51] [WARNING] [2.dnscrypt-cert.resolver1.dnscrypt.eu.] certificate is about to expire -- if you don't manage this server, tell the server operator about it
[2019-09-08 06:29:00] [CRITICAL] [2.dnscrypt-cert.ipredator.se.] certificate will expire today -- Switch to a different resolver as soon as possible

PeterDaveHello:
> `SERVER_ERROR` is usually a timeout. A query was sent to the server, but no reply was received.

Interesting, I have different dnscrypt-proxy instances in the same LAN network, the one on Raspberry Pi frequently got `SERVER_ERROR`, the one on my laptop works better than that.

Have checked both has enough resource, the latest versions(the recent 3 releases), so I don't know what's going on, doesn't seem to be network connectivity issue, sometimes it even got all `SERVER_ERROR`.

I wonder before I erase the whole system, is there something else I can do, as cloudflared and Pi-hole works very well on it, doesn't seem to be an environment issue either.

WusWus:
Hello, 

since today I have the problem that I can reach with DNScrypt-proxy websites only very difficult. 

I'm using 'Simple DNSCrypt' (v0.6.6) as GUI for dnscrypt under Windows 7 64bit. 
After having problems with version 2.0.25 today, I replaced it with the new version 2.0.26. 
Without DNSCrypt all websites are easily accessible. 

In the 'query. log' I always find 'SERVER_ERROR'. Duration to high? 
I only use selected servers (for example: 'geekdns-hk', 'jp.tiar.app', 'dnscrypt-04.adsnomore.io', 'arvind-io', 'dnscrypt-07.adsnomore.io', 'ibksturm', 'd0wn-tz-ns1', 'dnscrypt-03.adsnomore.io', 'soltysiak'), but only 'geekdns-hk' and 'opennic-ethservices' are available. 
All other servers are not reached/found. 

[...]

PR #971, opened 2019-10-19
differentiate between timeout and other error for dnscrypt servers

alisonatwork:
As per the suggestion of @PeterDaveHello in #924 this PR will print `SERVER_TIMEOUT` in the log if the error coming back from a DNSCrypt server is a net error with timeout flag.

I didn't change it for DoH because I had a look through the code and am not sure if a net error would get propagated back up to `proxy.go`.


jedisct1:
Thank you!

PR #973, opened 2019-10-20
More simple cleanups

maage:
Three style cleanups and another three changes.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -373,6 +373,7 @@ func (proxy *Proxy) processIncomingQuery(serverInfo *ServerInfo, clientProto str
 		return
 	}
 	pluginsState := NewPluginsState(proxy, clientProto, clientAddr, start)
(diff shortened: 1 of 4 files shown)


jedisct1:
👍

dnscrypt-proxy/proxy.go
```diff
@@ -417,7 +415,6 @@ func (proxy *Proxy) processIncomingQuery(serverInfo *ServerInfo, clientProto str
 					sharedKey, encryptedQuery, clientNonce, err := proxy.Encrypt(serverInfo, query, serverProto)
 					if err != nil {
 						pluginsState.returnCode = PluginsReturnCodeParseError
-						pluginsState.ApplyLoggingPlugins(&proxy.pluginsGlobals)
```
jedisct1, line 420: With the `defer`, that case may not be handled any more.

maage: Shouldn't it fail there?

jedisct1: Actually, yes, it should fail :)

PR #974, opened 2019-10-20
Expire xtransport cached IPs if resolved and add config_ttl for synthetic responses

maage:
Also check format of fallback_resolver

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -52,6 +52,8 @@ type Config struct {
 	CacheNegMaxTTL           uint32                     `toml:"cache_neg_max_ttl"`
 	CacheMinTTL              uint32                     `toml:"cache_min_ttl"`
 	CacheMaxTTL              uint32                     `toml:"cache_max_ttl"`
+	NegTTL                   uint32                     `toml:"neg_ttl"`
+	CloakTTL                 uint32                     `toml:"cloak_ttl"`
 	QueryLog                 QueryLogConfig             `toml:"query_log"`
 	NxLog                    NxLogConfig                `toml:"nx_log"`
 	BlockName                BlockNameConfig            `toml:"blacklist"`
@@ -103,6 +105,8 @@ func newConfig() Config {
 		CacheNegMaxTTL:           600,
 		CacheMinTTL:              60,
 		CacheMaxTTL:              86400,
+		NegTTL:                   600,
+		CloakTTL:                 600,
 		SourceRequireNoLog:       true,
 		SourceRequireNoFilter:    true,
 		SourceIPv4:               true,
@@ -365,6 +369,8 @@ func ConfigLoad(proxy *Proxy, svcFlag *string) error {
 
 	proxy.cacheMinTTL = config.CacheMinTTL
 	proxy.cacheMaxTTL = config.CacheMaxTTL
+	proxy.negTTL = config.NegTTL
+	proxy.cloakTTL = config.CloakTTL
 
 	proxy.queryMeta = config.QueryMeta
 
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


dnscrypt-proxy/xtransport.go
```diff
@@ -67,11 +77,38 @@ func NewXTransport() *XTransport {
...
+		item.ttl = time.Now().Add(ttl)
+	}
+	xTransport.cachedIPs.cache[host] = item
 	xTransport.cachedIPs.Unlock()
-	dlog.Info("IP cache cleared")
+}
+
+// If expire is true, remove data if expired
+func (xTransport *XTransport) loadCachedIP(host string, expire bool) (net.IP, bool) {
+	xTransport.cachedIPs.RLock()
```
jedisct1: I don't think a read lock is safe since we may remove entries.

dnscrypt-proxy/config.go
```diff
@@ -102,6 +104,7 @@ func newConfig() Config {
 		CacheNegMaxTTL:           600,
 		CacheMinTTL:              60,
 		CacheMaxTTL:              86400,
+		ConfigTTL:                600,
```
jedisct1: Having this knob is great!

But `configTTL` is not a very self-explanatory name. It doesn't expire a configuration :)
Can you think of a better name for it?

dnscrypt-proxy/xtransport.go
maage: Yes

dnscrypt-proxy/config.go
maage: I checked some others,

unbound has infra-host-ttl
dnsmasq has local-ttl, auth-ttl, neg-ttl

Maybe pluginsState.action == PluginsActionReject case:
refused_ttl / reject_ttl / blocked_ttl / neg_ttl

Add cloak_ttl for plugin_cloak. With default lik 600, it would solve all need adequately currently visible in example-cloaking-rules.txt.

And allow to set per record ttl in cloaking-rules.txt. As if you are testing you might want to use 0 and localhost could use like 86400 or greater. But that is another patch and I'll leave it to somebody else to implement when they have need to have something like that.

But anyways, most other caching dns servers default cache_min_ttl to 0 and recommend not to increase it. And it would be great to even provide some cache for blocked requests, or your log is then full of them.

maage:
I fixed wrong locking and renamed config_ttl -> neg_ttl and then added cloak_ttl.

dnscrypt-proxy/example-dnscrypt-proxy.toml
```diff
@@ -328,7 +331,9 @@ cache_neg_min_ttl = 60
 
 cache_neg_max_ttl = 600
 
+## TTL when dnscryp-proxy does reject entry
```
jedisct1, line 334: Typo :)

jedisct1:
There is a problem with expiring the server IPs.

The resolution happens way too frequently. I just tried with the `xfinity` resolver, and the proxy pauses every 30 seconds to refresh the server IP.

~~It also always uses the fallback resolver even though after startup, we have valid DNS settings (and can possibly get a more relevant IP since we ask the actual resolver instead of a fallback).~~ (I had `ignore_system_dns = true`, my bad).

~~The proxy also cannot start any more if no fallback were configured, even though a system resolver may be available.~~ (same).

jedisct1:
Weird.

With `ignore_system_dns = true`, I get:

```text
[2019-11-01 21:40:20] [DEBUG] Resolving [doh.xfinity.com] using resolver udp[9.9.9.9:53]
[2019-11-01 21:40:20] [DEBUG] [doh.xfinity.com] IP address [96.113.151.142] added to the cache, valid until 30s
```

But with `ignore_system_dns = false`:

```text
[2019-11-01 21:41:26] [DEBUG] [doh.xfinity.com] IP address [96.113.151.143] added to the cache, valid until 24h0m0s
```

jedisct1:
```go
        if ttl < xTransport.timeout {
            ttl = xTransport.timeout
        }
```

Maybe because we get a `CNAME`, the `ttl` is seen as `0`, and then set to the default `timeout` which is 30 seconds?

jedisct1:
The minimum TTL should be higher in order to minimize usage of the fallback resolver.

jedisct1:
My bad, Comcast records for their DoH server actually have a 30 sec TTL. This is silly. Other DoH servers aren't much better though.

PR #980, opened 2019-10-23
make sure app main goroutine quits, before stopping

vbauerster:
I've noticed that `*App.wg` never waited in main.go, so this PR fixes this.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -477,7 +477,7 @@ func ConfigLoad(proxy *Proxy, svcFlag *string) error {
 	}
 	proxy.showCerts = *showCerts || len(os.Getenv("SHOW_CERTS")) > 0
 	if proxy.showCerts {
-		proxy.listenAddresses = nil
+		proxy.listenAddresses = proxy.listenAddresses[0:0]
 	}
 	dlog.Noticef("dnscrypt-proxy %s", AppVersion)
 	if err := NetProbe(netprobeAddress, netprobeTimeout); err != nil {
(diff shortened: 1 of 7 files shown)


jedisct1:
Hi Vlad,

And thanks for your contribution!

I'd like to hold off for a bit before making internal changes, so that people can test the current beta, and version 2.0.29 (final, non beta) can be released soon.
So, for a couple days, only important bugs found in the beta should go in. Other changes will be made right after the final version is released.

Your patch looks good, although it probably needs a bit of testing on different platforms including Windows.

vbauerster:
My setup is `unbound` as front, backed by `dnscrypt-proxy` via systemd sockets. After some usage, I noticed that if I restart `dnscrypt-proxy` via `systemctl`, `unbound` gives empty results after a while, so I have to restart the whole system. Some code review revealed that connections were not closed properly (goroutine that defers Close() never quits). So I think this is the reason why `unbound` stops to respond if `dnscrypt-proxy` restarted.

Second commit tries to fix this, i.e. make sure all conn are closed on stop signal. I've tested with this patch and `unbound` continues to respond after `dnscrypt-proxy` have been restarted.

jedisct1:
Alright. It seems to work fine on Windows and macOS, so let's merge this.

Thanks again, Vlad!

jedisct1:
The downside of that change is that we can't quit any more while the servers list is being built. Which can take some time with the default settings (all resolvers).

vbauerster:
Ok, I think I got what you mean. I'll review this one more time.

PR #989, opened 2019-10-27
More robust source cache

### LINKED ISSUE #342: Unable [to] use source - Invalid signature?
Salut Frank.

I got this the other day on Windows power-up:

```
[2018-04-08 00:06:53] [INFO] Loading source information from URL [https://download.dnscrypt.info/resolvers-list/v2/public-resolvers.md]
[2018-04-08 00:06:53] [NOTICE] System DNS configuration not usable yet, exceptionally resolving [download.dnscrypt.info] using fallback resolver [9.9.9.9:53]
[2018-04-08 00:06:54] [CRITICAL] Unable use source [public-resolvers]: [Invalid signature]
[2018-04-08 00:06:54] [FATAL] No servers configured
```

Not sure where it got this invalid signature error from, corrupted download? It loaded fine 10 minutes later (time it took me to discover I had no DNS.) Wondering if some kind of retry setting should be added to DNSCrypt-Proxy. Or maybe I should just specify the same source twice? I don't know how often this will happen, wondering if I should start thinking about mitigating this.

Thanks.

### LINKED ISSUE #810: any solution/workaround if yesterday's invalid-signature error will repeat again?
wonna suggest additional parameter in dnscrypt-proxy.toml - "use previous public-resolvers.md in emergency case", which had to be set to NO with explanation (in capital letters too ;-) ), that setting this to YES would lead into an unbreakable vicious circle.

### LINKED ISSUE #854: refresh_delay is no longer being respected
dnscrypt-proxy is updating the **public-resolvers.md** and the **public-resolvers.md.minisig** on my Windows 10 device every 24 hours even though **refresh_delay** is set to 72(hours) in `dnscrypt-proxy.toml`

The **Date Modified** time stamp is getting updated every 24 hours as per the above image which shouldn't happen every 24 hours but instead every 72 hours. I believe this is a regression as this didn't happen before.

### LINKED ISSUE #988: relays.md.minisig invalid signature
I'm getting an invalid signature when I start dnscrypt-proxy.
I found out in the repository that relay.md was modified while relay.md.minisig was not.

Imgur

I was somehow able to temporarily fix it by downloading an old relay.md.

**P.S.** Maybe in a future release, it would be helpful if old .md and .minisig are renamed to .bak when being updated in case of invalid signatures.

welwood08:
The following issues should be fixed by this PR:
- When a signature fails to validate, the cache is no longer destroyed as a side-effect.
- The 2 files that make up each source are treated as a pair in failure handling, meaning a valid cache should not be made invalid by a single bad download.
- Prefetch downloads now use the same code path as startup downloads, meaning:
  - signatures will always be validated before caching to disk,
  - all configured URLs are tried (not just the first),
- If a downloaded source's signature fails to validate then the cache will be used as a fall back (even if it's old).
- Duplicated (and outdated) logic to handle webserver errors removed (XTransport already does it).
- When a prefetch download fails, try again in 10mins instead of 1min (not too important, but I noticed this made my debug logs even more spammy than they should be).

I pulled in a dependency on `github.com/powerman/check` to make test output easier to read, at the cost of quite a few more sub-deps being pulled in.

I've not yet done any long-running testing with full builds using this code, but if there's no hurry to merge this then I plan to test on win64, freebsd/amd64, linux/mips64 and linux/amd64.

I'm also unsure if this has any negative effects on @vbauerster's recent fixes to quit behaviour since I touched some of the same code - I'd appreciate any insight and/or testing in that area in particular, especially if anything got lost when I rebased.

Partially fixes #342 - looks the idea of retrying downloads at startup would need support for runtime (re)configuration, which is a task for some other day.
Fixes #810
Fixes #988

```console
$ go test -cover
PASS
coverage: 5.8% of statements
  checks:  1686 passed  0 todo  0 failed	(total)
ok  	github.com/jedisct1/dnscrypt-proxy/dnscrypt-proxy	0.258s
```


jedisct1:
Thank you! This is great, and will definitely be very useful.

Just one thing: we may want to keep `MinSourcesUpdateDelay` (or something else, but not the `refresh_delay` value).

`refresh_delay` is essentially how old we accept a list to be, i.e. how long a computer can be turned off and turned on again without having to download something at startup.

The purpose of background preloading is to keep the lists fresh once the network can be used, so that `refresh_delay` works as expected. If that process happens less frequently, people with extreme firewall rules can have issues right after a restart.

welwood08:
> Just one thing: we may want to keep `MinSourcesUpdateDelay` (or something else, but not the `refresh_delay` value).
> 
> `refresh_delay` is essentially how old we accept a list to be, i.e. how long a computer can be turned off and turned on again without having to download something at startup.
> 
> The purpose of background preloading is to keep the lists fresh once the network can be used, so that `refresh_delay` works as expected. If that process happens less frequently, people with extreme firewall rules can have issues right after a restart.

Thanks for the clarification of the intent, I think the name was causing confusion so I've separated the 2 concepts internally:
- `cacheTTL`: "how old can the cache be at startup before trying to download an update immediately"
- `prefetchDelay`: "how long should the background prefetcher wait between successful downloads"

The former is `refresh_delay` in the configuration, the latter is the constant `DefaultPrefetchDelay` (previously known as `MinSourcesUpdateDelay`). I'll leave it up to you if you want to change the name of `refresh_delay` in the configuration to clarify its purpose to users or make `DefaultPrefetchDelay` configurable (which should be trivial now).

Either way, this PR should make extreme firewall startup failures less common since download problems are no longer fatal if there's a correctly-signed cache, regardless of how old it is.

jedisct1:
Excellent!

Yes, renaming these makes everything way less confusing. Thanks for doing this!

We can keep `refresh_delay` in the configuration file to avoid breaking backwards compatibility.

jedisct1:
Do you think this is good to be merged?

welwood08:
I'm close to pushing a re-worked update to this branch with separated commits, once that's done I'll hit the ready for review button to take it out of draft status. I'd still want to do my own longer-term testing, but I'll let you decide how quickly you want to merge it since it's quite a large code change.

jedisct1:
Fantastic! Yes, separated commits would be better.

Tests are badly needed, but mocking servers would be a huge task.

I'd rather be inclined to just do system testing instead. We can easily spin up a local DoH and DNSCrypt server, and send actual queries through the proxy to see if we get the expected results using multiple configuration files. That would be useful to simultaneously test the server implementations.

welwood08:
I agree that there's clearly been a need recently for basic sanity checks like making sure the proxy starts and can respond to queries on at least some of the supported platforms. I had been wondering how I could help with that, but couldn't start anything with this massive rebase in progress. I don't want to get off topic for this PR, so maybe implementation ideas could be collected in a new issue?

jedisct1:
It doesn't even have to be in Go. It can be scripts that we run on CI. If Travis is too slow for this, we can run them as Github actions or on Azure Pipelines in parallel.

welwood08:
In case you're wondering, after I force-pushed my completed PR and hit the ready button I realised I hadn't rebased onto latest master. I hit a few unexpected bumps in that task where bits of code I had relied on disappeared or changed, so still not ready to merge but much closer.

welwood08:
OK, I think I'm done force-pushing for now @jedisct1 if you want to do any review before merging. I'm moving on to real-world testing on the platforms I mentioned above but don't let that affect your review/merge timeline, I can fix any bugs I notice on master if it's already merged.

welwood08:
Fixed conflicts with master in 1st 2 commits and force pushed branch again.

alisonatwork:
I love that we are adding some testability here, but unfortunately the tests do not run when I try to compile on Windows:

```
$ uname
MINGW64_NT-10.0-18362
$ go test
# github.com/powerman/check
..\..\..\go\pkg\mod\github.com\powerman\check@v1.2.0\color.go:32:12: undefined: unix.IoctlGetTermios
FAIL    github.com/jedisct1/dnscrypt-proxy/dnscrypt-proxy [build failed]
$
```

I tried commenting out all the UNIX stuff in `color.go`, which got the tests to run, but most of them then failed due to a newline/carriage return problem. I manually changed all the line endings to UNIX endings, but we could force this by putting a `.gitattributes` to set the newlines of the fixtures in `test_data` directory.

After that, I still got 8 test failures. They're all the same assertion failure - here is one:

```
    --- FAIL: TestNewSource/cache_open-err,_download_no-urls/7-0 (0.00s)
        sources_test.go:353: Unexpected error
            Checker:  Match
            Regex:    (string) (len=17) 'permission denied'
            Actual:   (string) (len=28) 'Incomplete encoded signature'
```

I don't have time to look into that one right now unfortunately.

alisonatwork:
FYI for the ioctl thing i just opened https://github.com/powerman/check/pull/3

jedisct1:
Yeah, we shouldn't base any tests on specific error messages, even for a given platform.

welwood08:
Thanks @alisonatwork for taking the time to run the tests on Windows (I hadn't even thought to run them outside my dev env) and for the upstream patch! I like the `.gitattributes` solution you suggest since it's Git's fault the line endings changed, I'll amend the commit that introduces the fixtures.

> Yeah, we shouldn't base any tests on specific error messages, even for a given platform.

@jedisct1 I disagree, a different error message indicates that an unexpected code path was followed and the test isn't testing what it thinks it's testing.

In this particular case, the test is meant to simulate how the code handles an error when opening a cache file and in order to do this it writes a cache file that has no read permission bits. As it turns out, this doesn't work on Windows and the tested code is able to read the cache file and then check its signature. While it still encounters an error and the effect is the same, it indicates a fault in the test which would have remained undiscovered if it wasn't looking for a specific message.

welwood08:
All the problems @alisonatwork noticed should be resolved now. I needed to set `core.autocrlf=true` in my Windows .gitconfig to reproduce the line endings problem (it had been set to `input`, presumably because I encountered something similar in the past). Unit tests are successful on Linux, FreeBSD and Windows for me.

alisonatwork:
Sorry i didn't get to test this before it was merged, my VPN didn't let me get out of Great Firewall so i couldn't download the deps. Now it's back, looking good. Thanks all 👍

jedisct1:
Thank you Alison!

Issue #976, opened 2019-10-21
Windows service is killed by event 7000 if there is no network on startup


alisonatwork:
My laptop usually doesn't have internet connection when I start it up, and sometimes it can be 20 minutes or so till I connect it to the wifi. This should be handled by the `netprobe_timeout` setting - by putting it to `-1` it should retry up to 1 hour until the network appears. However, this does not work.

This is what appears in the event log:

```
由于下列错误，dnscrypt-proxy 服务启动失败: 
服务没有及时响应启动或控制请求。
等待 dnscrypt-proxy 服务的连接超时(45000 毫秒)。
``` 

Sorry my Windows is Chinese, but this is events 7000 and 7009: https://support.microsoft.com/zh-cn/help/922918/a-service-does-not-start-and-events-7000-and-7011-are-logged-in-window

If a service doesn't respond within 30-45 seconds, the Windows service manager will just kill it.

A temporary workaround might be to install the service with a retry behavior, so Windows will restart it after it dies, but I think the right fix here is to move the netprobe into the main loop of the service instead of making it happen while loading the config file on startup.

Moving this code into the main loop might break startup behavior on other OSes, however. And it will definitely affect the startup commands like `list`, `check` and so on.

I don't really have time to look into this right now, so opening a ticket instead.

jedisct1:
Moving the netprobe doesn't look like a good fix. Loading the lists or benchmarking the servers may also exceed the service deadline.

Ideally, the service should return immediately, and everything will then happen in the main loop.

I once tried to reorganize the code and `kardanios/service` to do this, but couldn't get something that didn't break either systemd or the windows service manager.

alisonatwork:
Okay, I've made an attempt to address this in PR #991 

I am not sure of how it will impact systemd. If it's using the PID file, then it shouldn't be any different from before, because the PID file doesn't get created until `appMain`. If it's not using the PID file... then that's beyond my basic systemd knowledge 🤣 But looking at the source code of kardianos/service, it doesn't seem like there is any special state that is maintained in the `Run` loop for systemd, so I think it should be okay.

alisonatwork:
More interesting reading if anyone searches this thread and wants to know more about how Windows service state transitions: https://docs.microsoft.com/en-us/windows/win32/services/service-status-transitions

jedisct1:
So, this is fixed :) Thank you Alison! That was a long-standing issue.

PR #991, opened 2019-10-27
move ConfigLoad into Start function when running as a service

alisonatwork:
This is an attempt to fix the issue I mentioned in #976 - where the Windows service manager kills the service before it starts up, if it didn't initialize in time.

The recommendation of other people encountering this issue with kardianos/service is simply to make sure the `Start` method returns as soon as possible by doing all the init async: https://github.com/kardianos/service/issues/131

So that is what I am doing here. If there is a `svc` object (i.e. running as a service), then `ConfigLoad`  runs inside a goroutine in `Start` - the same one that previously only ran `appMain`. If there is no `svc` object, then the `ConfigLoad` just happens in the main thread before setting up the signal listener and calling `appMain`.

This does mean that the config is no longer loaded before the block that handles service control (`len(*svcFlag) != 0`). So if your config is broken it will now still install the service. My guess is that most users will install the service with a default or blessed config first (i.e. through package manager or installer), then if they want to customize their own config they will test it with `-check` or `-list` before restarting (or reinstalling), so this shouldn't impact most cases. I think this simplifies the code a bit because there no longer needs to be a check in the config loader to see what the command-line service arguments were.

I am currently running a test version of this on my Windows and it seems to work okay. I could compile one inside a Windows Subsystem for Linux, but it would probably be better if someone with a real Linux could test this in systemd. I can't see it causing any major problems, since any failure in the config file drops a `dlog.Fatal`, which does `os.Exit(255)`, and there is no PID file written until `appMain` is invoked anyway.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -211,7 +211,7 @@ func findConfigFile(configFile *string) (string, error) {
 	return path.Join(pwd, *configFile), nil
 }
 
-func ConfigLoad(proxy *Proxy, svcFlag *string) error {
+func ConfigLoad(proxy *Proxy) error {
 	version := flag.Bool("version", false, "print current proxy version")
 	resolve := flag.String("resolve", "", "resolve a name using system libraries")
 	list := flag.Bool("list", false, "print the list of available resolvers for the enabled filters")
@@ -225,9 +225,6 @@ func ConfigLoad(proxy *Proxy, svcFlag *string) error {
 
 	flag.Parse()
 
-	if *svcFlag == "stop" || *svcFlag == "uninstall" {
-		return nil
-	}
 	if *version {
 		fmt.Println(AppVersion)
 		os.Exit(0)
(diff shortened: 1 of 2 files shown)


jedisct1:
This is very, very, very cool, thank you for tackling this, Alison!

Since it requires a bit of testing on different platforms, do you mind if we wait until 2.0.29 has been tagged before merging this?

alisonatwork:
It sounds very reasonable to wait till 2.0.29 is out the door, for sure 👍

jedisct1:
Hi Alison,

Looks like it breaks all the `-service` commands :( They are now ignored.

Can you take a look?

The change has been temporarily reverted.

alisonatwork:
Argh! Sorry about that. I think the problem is this line inside `ConfigLoad`: `flag.Parse()`. I think until `flag.Parse()` is called, `svcFlag` is never allocated, so the `-service` code branch in `main()` does not work properly.

I think that the flags should be parsed before calling `ConfigLoad`, then put into some kind of "parsed flags" container which can be used when `ConfigLoad` is finally called. I don't have time to make that fix right now, but I can give it a look when I am home from work, or worst case over the weekend.

Apologies again! 🙈

PR #997, opened 2019-10-30
move ConfigLoad into Start function when running as a service (take 2)

alisonatwork:
This is a second attempt at fixing the problem detailed in #976 where loading the config can take too long and then Windows kills the service before it's started. The first attempt was in #991 but it had a flaw that it broke the `-service` argument.

Here is my new version. What I have done is moved the flags parsing out of `config.go` and into `main.go`. By doing the flags parsing on the top level, we are able to process all of the "fast" flags (e.g. `-version`) immediately, then exit.

I have added a new struct called `ConfigFlags` that saves all the pointers to the parsed flags that need to be used in `ConfigLoad`. If we are running as a service, we store a pointer to that struct on the `App`, then pass it into `ConfigLoad` during the `Start` function of the service. If we are not running as a service, then we just pass the pointer direct.

A different implementation that I considered was to store all of the flags directly into `App` from the outset. This way no matter where we are in the application, we should always be able to get the parsed flags. It also means there would be one source of truth for all the parsed flags, instead of having some of them parsed into `ConfigFlags` and other ones used directly in `main()`. The down-side is that it would create a circular dependency from `main.go` to `config.go` (to load the config) and back (to find the application flags type). That might be okay in Golang, but it feels a little odd to me. Still, I would be happy to refactor it to work this way if seems like a better option?

Either way, I definitely have tested the `-service` flag now and this does seem to work correctly. Apologies for the bug on the first PR.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -194,6 +194,17 @@ type ServerSummary struct {
 	Stamp       string   `json:"stamp"`
 }
 
+type ConfigFlags struct {
+	List                    *bool
+	ListAll                 *bool
+	JsonOutput              *bool
+	Check                   *bool
+	ConfigFile              *string
+	Child                   *bool
+	NetprobeTimeoutOverride *int
+	ShowCerts               *bool
+}
+
 func findConfigFile(configFile *string) (string, error) {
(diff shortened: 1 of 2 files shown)


jedisct1:
Looks great!

Thank you so much, Alison! You rock!

PR #1001, opened 2019-10-31
check error that was being erroneously shadowed

ericlagergren:


welwood08:
I initially assumed the same thing was happening 10 lines above, but once I started to understand what the code was doing I began to suspect this might fix a bug I had seen where downstream DNSSEC validation would gradually break one domain at a time over several days and I'd need to restart the service. Is this what prompted your fix or were you just reading through the code?

Since it wasn't immediately clear what the consequences of shadowing were here, if I understand correctly the `if err != nil` is fine either way but after that an error returned from `response, err = proxy.exchangeWithTCPServer(...)` is thrown away on leaving the scope on the next line.

LGTM at least, but since Go's still new to me and this gets close to some cryptography I'll leave the final decision to @jedisct1.

ericlagergren:
I ran staticcheck (static analysis tool) and it complained that err was assigned to, but unused.

The consequence of the shadowing is that the error will never be checked.

Capture d’écran 2019-10-30 à 8 47 27 PM

jedisct1:
Not a big deal since it only changes the response error code, but yep, good catch, `err` shouldn't be shadowed here. 

Thanks!

PR #1003, opened 2019-10-31
Pull980fix

vbauerster:
follow up on [comment](https://github.com/DNSCrypt/dnscrypt-proxy/pull/980#issuecomment-548153169).

### DIFF
--- a/dnscrypt-proxy/main.go
+++ b/dnscrypt-proxy/main.go
@@ -109,15 +109,7 @@ func main() {
 }
 
 func (app *App) Start(service service.Service) error {
-	go func() {
-		if err := ConfigLoad(app.proxy, app.flags); err != nil {
-			dlog.Fatal(err)
-		}
-		if err := app.proxy.InitPluginsGlobals(); err != nil {
-			dlog.Fatal(err)
-		}
-		app.appMain()
-	}()
+	go app.appMain()
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


jedisct1:
Thank you!

vbauerster:
No prob, thank you for review!

jedisct1:
Oops, not good.

It doesn't start any more on BSD.

```sh
$ ./dnscrypt-proxy
[2019-10-31 16:58:29] [DEBUG] No service system detected.
[2019-10-31 16:58:29] [NOTICE] Stopped.
```

jedisct1:
Seems to be broken on Android as well.

jedisct1:
`app.Start(nil)` is called, but is not blocking, so the app immediately ends.

jedisct1:
Replacing `app.Start(nil)` with `app.appMain()` make the application usable, but it now requires two signals to quit it :/

vbauerster:
I have a quick fix, wait!

vbauerster:
I hope following patch should fix this, but not sure if I should add commit to this PR or make a new one.
[](https://postimg.cc/LYPKWwMB)

I suppose `pidfile` is created for android and bsd, so calling `app.Stop(nil)` in any case.

jedisct1:
Version 2.0.29-beta.3 was pretty stable, so I tagged a new version that's essentially 2.0.29-beta.3. 

Merging too many changes between the beta and the final version wasn't a good idea at all, my bad :(

Now that 2.0.31 should be a stable version that supports anonymization, we can relax and take a bit more time to test tricky changes like this :)

vbauerster:
ok, let me know if you still interested in this PR.

PR #1007, opened 2019-10-31
Move ConfigLoad into AppMain

alisonatwork:
This is the third attempt of the change previously found in #991 and #997 - it seems my changes conflicted with #980 and some other startup alterations.

I think this should apply clean, however. The first commit is same as before - move the flag parsing into `main.go` (from `config.go`) - this is a pure refactoring and should be harmless. The second commit just moves `ConfigLoad` and `InitPluginsGlobal` into `AppMain`. Since the service runs `AppMain` in a goroutine and the non-service runs it in the foreground, it does not matter whether the those init pieces happen in `Start` or `AppMain`.

I believe this change is orthogonal to the other one, which seems to have more to do with cleaning up after exit. I'm happy to hold off on this and try a different approach if it makes more sense for the other one to go first, though.

### DIFF
--- a/dnscrypt-proxy/main.go
+++ b/dnscrypt-proxy/main.go
@@ -23,6 +23,7 @@ type App struct {
 	wg    sync.WaitGroup
 	quit  chan struct{}
 	proxy *Proxy
+	flags *ConfigFlags
 }
 
 func main() {
@@ -67,7 +68,9 @@ func main() {
 		os.Exit(0)
 	}
 
-	app := &App{}
+	app := &App{
+		flags: &flags,
+	}
 	svc, err := service.New(app, svcConfig)
 	if err != nil {
 		svc = nil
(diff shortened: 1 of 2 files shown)


dnscrypt-proxy/main.go
```diff
@@ -101,6 +119,14 @@ func (app *App) Start(service service.Service) error {
...
 
 func (app *App) AppMain() {
+	if err := ConfigLoad(app.proxy, app.flags); err != nil {
+		dlog.Fatal(err)
+	}
+	if err := app.proxy.InitPluginsGlobals(); err != nil {
+		dlog.Fatal(err)
+	}
+	app.quit = make(chan struct{})
+	app.wg.Add(1)
```
alisonatwork, line 129: This is the only strange piece left in this PR. We have a waitgroup and quit channel sitting here that aren't really necessary. But i presume these will be used at some point in a future PR to manage a clean shutdown.

jedisct1: I don't even remember what the `quit` channel is for :) Probably not necessary indeed.

The waitgroup is indeed for a clean shutdown. It's not really used here, but will with @vbauerster changes.

PR #1010, opened 2019-11-01
Minor cleanups

alisonatwork:
In the spirit of recent cleanups, here I remove some more unused code - return values always ignored or errors that were never set.


jedisct1:
👍

PR #1028, opened 2019-11-08
set go.mod to unix endings

alisonatwork:
Now that we have a gitattributes, I'd like to add `go.mod` to it, since running a `go build` on Windows causes this file to get re-written with different line endings.


jedisct1:
Of course!

PR #1068, opened 2019-11-24
plugin_block_name: also block cname

### LINKED ISSUE #1067: Feature Request: CName Uncloaking
Following several news regarding CName Cloaking which is very privacy invasive and has several security implications, it seems interesting to implement block on CName. 

**TL;DR:** CNAME can be used to masquerade third-party trackers as first-party trackers. DNS is not exposed to addons like UBlock Origin meaning it is hard to block ads and tracker using CNAME. As they are now "first-party", they often get cookies from the main domain (Analytics ID, Auth, CSRF, ...) which can be used to get even more information than ever. See references for more information.

**Expected behaviour:**
Domain and CNAME are blocked the same way

If I am using the following blacklist:
```
*.eulerian.net
```
I expect:
```
$ dig f7ds.liberation.fr
;; ANSWER SECTION:
f7ds.liberation.fr.	1	IN	HINFO	"This query has been locally blocked" "by dnscrypt-proxy"

$ dig liberation.eulerian.net.
;; ANSWER SECTION:
liberation.eulerian.net.	1	IN	HINFO	"This query has been locally blocked" "by dnscrypt-proxy"
```

**Current behaviour:**
Domain is blocked but resolved if in a CNAME, this means that CNAME can be used to bypass blacklist or any specific rule
```
$ dig f7ds.liberation.fr
;; ANSWER SECTION:
f7ds.liberation.fr.	3494	IN	CNAME	liberation.eulerian.net.
liberation.eulerian.net. 3494	IN	CNAME	atc.eulerian.net.
atc.eulerian.net.	3494	IN	A	109.232.197.179

$ dig liberation.eulerian.net.
;; ANSWER SECTION:
liberation.eulerian.net. 1	IN	HINFO	"This query has been locally blocked" "by dnscrypt-proxy"

```

**References:**
https://news.ycombinator.com/item?id=21604825
https://news.ycombinator.com/item?id=21610386

Sayrus:
```
Currently, CNAME can be used to resolve blocked domain (Just CNAME an
allowed domain to a blocked one and the server often return both the
CNAME and the A or AAAA field if it exists).

This is a naive implementation that check if an answer exists, check
every CNAME field in it against the blacklist.
It stops on the first error or on the first blocked CNAME.
```

I am not proficient in Go which is why I do think this implementation is not that great. However, I think a reference might be helpful.

**Note:** As this PR adds PluginBlockName as a response plugin, it might make sense to avoid checking responses with NXDomain and other errors against the blacklist.

Provides a sample implementation and close #1067

### DIFF
--- a/dnscrypt-proxy/plugin_block_name.go
+++ b/dnscrypt-proxy/plugin_block_name.go
@@ -92,8 +92,22 @@ func (plugin *PluginBlockName) Eval(pluginsState *PluginsState, msg *dns.Msg) er
 	if plugin.blockedNames == nil || pluginsState.sessionData["whitelisted"] != nil {
 		return nil
 	}
+	answers := msg.Answer
 	questions := msg.Question
 	if len(questions) != 1 {
+		return nil
+	}
+	if len(answers) > 0 {
+		for _, answer := range msg.Answer {
+			header := answer.Header()
+			Rrtype := header.Rrtype
+			if Rrtype == dns.TypeCNAME {
+				status := plugin.blockedNames.check(pluginsState, answer.(*dns.CNAME).Target)
+				if (status != nil || pluginsState.action == PluginsActionReject) {
+					return status
+				}
+			}
+		}
 		return nil
 	}
 	return plugin.blockedNames.check(pluginsState, questions[0].Name)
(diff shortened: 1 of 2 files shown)


jedisct1:
Thanks! This has been added.

PR #1102, opened 2019-12-09
Add multiple error checks

milgradesec:


jedisct1:
Maybe we can call `dlog.Crit()`, but if we can't log something, maybe we shouldn't prevent a DNS query from stopping being processed and getting a response.
Same for `Write`s to the client. It they fail because the client disconnected, maybe we shouldn't abort the plugin chain. The response we got can still be cached and logged. If the client didn't get it, it's no big deal.

PR #1120, opened 2019-12-21
add some nonexistent zones

hugepants:
Add some more nonexistent zones for the block_undelegated feature.


jedisct1:
Thank you!

Do you see `.localhost` as a zone? Not just `localhost`?

hugepants:
I don't remember - sorry. I've since deleted my logs!
I've enabled logging again so I'll let you know if I spot anything.
Thanks

PR #1137, opened 2020-01-06
.

ghost:
.


r4sas (approved):

jedisct1:
I’m against this.

With Unicode strings, `\d` is locale dependant. It is not equivalent to `0-9`, but matches anything that represents a digit for the user’s locale. This is slow, non-deterministic and potentially dangerous.

And even if we override the configured locale, or force the string to not be a Unicode one, `\d` is semantically incorrect. We don’t want a digit here. We want something in the `0-9` set.

PR #1176, opened 2020-02-03
WIP: add openbsd pledge and unveil support

canarysnort01:
I've been tinkering with adding both pledge and unveil to dnscrypt-proxy.
It's not working completely, but I thought I would share my progress.

This currently requires a patch to the openbsd kernel because it refuses
to allow the kern.somaxconn sysctl which go looks up when it listens on
a tcp socket[1][2][3].

This breaks non-main log files and pid files as well, because
os.MkdirAll gets confused when unveil is used. I'm not sure how to
resolve this.

[1] https://golang.org/src/net/sock_posix.go#L57
[2] https://golang.org/src/net/net.go#L373
[3] https://golang.org/src/net/sock_bsd.go#L27

OpenBSD kernel patch:

```
diff --git sys/kern/kern_pledge.c sys/kern/kern_pledge.c
index 9f436df4893..8d1203198ed 100644
--- sys/kern/kern_pledge.c
+++ sys/kern/kern_pledge.c
@@ -904,6 +904,12 @@ pledge_sysctl(struct proc *p, int miblen, int *mib, void *new)
                        return (0);
        }

+       if ((p->p_p->ps_pledge & PLEDGE_INET)) {
+               if (miblen == 2 &&              /* kern.somaxconn */
+                   mib[0] == CTL_KERN && mib[1] == KERN_SOMAXCONN)
+                       return (0);
+       }
+
        if ((p->p_p->ps_pledge & (PLEDGE_ROUTE | PLEDGE_INET | PLEDGE_DNS))) {
                if (miblen == 6 &&              /* getifaddrs() */
                    mib[0] == CTL_NET && mib[1] == PF_ROUTE &&
```

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -242,7 +242,7 @@ func findConfigFile(configFile *string) (string, error) {
 	return path.Join(pwd, *configFile), nil
 }
 
-func ConfigLoad(proxy *Proxy, flags *ConfigFlags) error {
+func ConfigRead(flags *ConfigFlags) (*Config, error) {
 	foundConfigFile, err := findConfigFile(flags.ConfigFile)
 	if err != nil {
 		dlog.Fatalf("Unable to load the configuration file [%s] -- Maybe use the -config command-line switch?", *flags.ConfigFile)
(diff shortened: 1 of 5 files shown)


jedisct1:
We can add `pledge(2)` support, but `unveil()` adds too much complexity here.

It is very unlikely that I will always remember to update/add/remove `unveil()` calls every time a  change requiring file access is made. And the current, incomplete patch already shows that even dealing with the current set of features is tricky and fragile.

If you are willing to maintain a fork, forever, that leverages `unveil(2)`, and use it to add patches to the OpenBSD port, go ahead. But I'd rather not merge something that will make the project very difficult to maintain, especially since Golang already offers some security guarantees.

canarysnort01:
I don't intend to officially maintain support for this if you're not interested in allowing it upstream. However, I will probably keep these changes for myself and publish what works for me on my github if anyone else is interested.

FYI OpenBSD has accepted the patch I posted above.

MystesofEternity:
@canarysnort01 Thank you very much for this! I simply can't assert enough that one does not simply have "too much" security!

Seems like I'll be using openBSD for a hardened local DNS server soon :D

PR #1184, opened 2020-02-08
Improve generate-domains-blacklist.py to remove redundant lines

### LINKED ISSUE #1183: generate-domains-blacklist.py not removing redundant lines
## *Who* is the bug affecting?
People using regular expressions in the domains-blacklist-local-additions.txt file for the generate-domains-blacklist.py script

Performance of dnscrypt-proxy and file size of the blacklist I suppose

## *When* does this occur?
If the domains-blacklist-local-additions.txt file includes expressions like `*analytic*` the resulting blacklist still contains entries such as google-analytics.com or analytics.google.com, even though the `*analytic*` expression would already take care of the blocking.
While that particular example in my case only causes 326 redundant lines, there are much worse cases, like `ads?.*` (1764 redundant lines) or similar that make up a lot of blacklist entries.

## *Where* does it happen?
Somewhere in generate-domains-blacklist.py, probably this line `if has_suffix(all_names, name) or name in unique_names:` in the process blacklists loop.

1. Add (for example) `*a*` to your domains-blacklist-local-additions.txt file
2. Run generate-domains-blacklist.py.
3. See that the resulting blacklist is still full of entries containing the letter a.

In that case no lines containing the letter a, except for the regular expression itself should appear.

Or in general, remove all lines that are already covered by regular expressions in the resulting blacklist.
This could either be done in the comparison or in another loop at the end, although the regular expressions might have to be converted (`*` becomes `.*`, . becomes `\.` and so on), depending on how python works.

## Other Comments
I can try fixing it myself, but I have no experience with python and it might not be the most efficient way

Huhni:
fixes #1183

### DIFF
--- a/utils/generate-domains-blacklists/generate-domains-blacklist.py
+++ b/utils/generate-domains-blacklists/generate-domains-blacklist.py
@@ -81,6 +81,32 @@ def parse_list(content, trusted=False):
     return names, time_restrictions
 
 
+# basic check if the line contains any regex specific char
+def is_regex(line):
+    regex_chars = "*[]?}{"
+    return any(char in line for char in regex_chars)
+
+
+def parse_regex(names):
+    regexes = set()
+    for line in names:
+        # skip lines without regex characters:
+        if not is_regex(line):
+            continue
+        # convert to python regex:
+        line=line.replace(".", "\.")
+        line=line.replace("*", ".*")
+        line = "^"+line+"$"
+        # check if resulting regex is valid:
+        try:
+            if re.compile(line):
+                regexes.add(line)
+        except re.error:
+            sys.stderr.write("Invalid regex: {} [{}]\n".format(line, re.error))
+            continue
+    return regexes
+
+
 def print_restricted_name(name, time_restrictions):
     if name in time_restrictions:
         print("{}\t{}".format(name, time_restrictions[name]))
@@ -134,6 +160,18 @@ def has_suffix(names, name):
     return False
 
 
+# check if a line matches with any of the collected regexes:
+def covered_by_regex(line, regexes):
+
+    # only check lines that aren't regexes themselves:
+    if not is_regex(line):
+        for regex in regexes:
+            if re.match(regex, line):
+                return True
+
+    return False
+
+
 def whitelist_from_url(url):
     if not url:
         return set()
@@ -148,6 +186,7 @@ def blacklists_from_config_file(
 ):
     blacklists = {}
     whitelisted_names = set()
+    all_regexes = set()
     all_names = set()
     unique_names = set()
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks!

Rules matched by dnscrypt-proxy are not regular expressions, so we probably don't want to assume they are in the Python script.

We should do strict matching of lines starting with `=` and do globbing (not regex matching. `example*.com` matches anything after `example` not `e` zero or more times as in a regex).

Also, these rules are only relevant in local ("trusted") sources.

Huhni:
Thank you, I didn't know that it was called globbing and thought it was just a different kind of flavor for regular expressions. With `import glob` it should be easier to implement without having to convert the glob expressions into regular expressions first and so on.
And yes it would make more sense to only look for globs in the local files, although maybe there are filter lists available online that also support globs 🤔 
But yeah I'll try to optimize the script again if I find the time for it.

Also another thing I noticed and wasted a **lot** of time on, was that the blacklist (and probably all configuration files) do not support UTF16, but powershell seems to use it as default encoding when redirecting output into a file with `>` like `python generate-domains-blacklist.py > list.txt.tmp`
I thought there was some invalid line or something and spent so much time looking for the reason until I finally figured it out 🐔 
Not sure if I should open a new issue for that or if adding a comment to the script that says something like 
> when using powershell use ` | out-file list.txt.tmp -encoding utf8` instead of `> list.txt.tmp`

would suffice

jedisct1:
Wow, good to know for Powershell using UTF16.

This is definitely something worth documenting. Or maybe we can add a `-o <file>` command-line switch to that the Python script directly writes the file.

Huhni:
I just noticed that `import globs` is only for actual filenames, but there is `fnmatch`, which pretty much also just converts a glob expression into a regular expression. I don't think there is anything for the strict matching with = though and I'm not sure what to do with it.
If say `=example.com` is in the domains-blacklist-local-additions.txt does that mean all other subdomains of example.com should be ignored? What if there are conflicts with other rules, say `ads.example.com`?
The changes I made so far only clean up the generated blacklist a little and don't change the logic behind the filtering, so I'm not sure how to implement that.

Also I don't know if I should extend the cleaning up part to the whitelist as well or if the domains-whitelist.txt doesn't support globs anyways. I'm getting a bit confused by the current code and I only use the actual whitelist.txt myself for some exceptions for some broader rules myself and not the domains-whitelist.txt for the generator script.

kamran1988 (approved):

lenwar:
This version of the script gives me the following error:

`$ python generate-domains-blacklist.py -o list.txt.tmp && mv -f list.txt.tmp list
  File "generate-domains-blacklist.py", line 230
    print("########## Blacklist from {} ##########\n".format(url), file=output)
                                                                       ^
SyntaxError: invalid syntax
`

The "^" marker is under the equals-sign of "file=output)"

In case it's related. I have the following Python versions:
Python 2.7.5 (CentOS 7)
and
Python 2.7.3 (Ubiquity USG)

PR #1192, opened 2020-02-14
A few tiny doc fixes

alisonatwork:
Just a few spelling and grammar tweaks.

Interesting one is "more than one resolver is" and "more than one resolver are", i confused myself looking at it, so i did a search 🤦‍♀   https://www.grammarphobia.com/blog/2007/12/can-more-than-one-be-singular.html


welwood08:
LGTM :+1:

PR #1202, opened 2020-02-23
Added DNS RRs filter plugin

### LINKED ISSUE #1201: DNS Resource Records filters
As HE (https://tunnelbroker.net/) IPv6 broker user i have a problem using certain services (ex. Netfix) because they are blocking access "due to proxy or VPN usage". 

As dnscrypt-proxy user i would like to filter AAAA DNS RR of Netflix (and others) domains.

Right now i`m using custom python Twisted DNS server that acts as primary DNS for my LAN. This "proxy" is filtering unwanted RRs and then calls dnscrypt-proxy. Problems with this solution are:
1. Python is slow
2. I`m using different DNS "entry-points" for normal and DOH DNS

This feature will benefit all users with special needs for filtering specific RRs. 

I`ve made a PR with new plugin, that acts similar to current "pattern-based query blocking", but for generic RR filtering.

YakhontovYaroslav:
Fixes #1201
This PR adds new plugin, that works similar to current patttern-based blocker, except this one can block specific DNS  RRs.


jedisct1:
From a functional perspective, this is doing exactly the same thing as the name blocker.

Having a distinct configuration file doesn't seem to be very convenient. Although it may require more code changes, such a feature may be easier to use if the record type was just an optional parameter for regular rules.

For example `*.example.com<TXT>`  would only match `TXT` records.

YakhontovYaroslav:
I have considered extending existing block_name plugin, but in the end changed my mind, due to potential breaking changes that will be introduced during development. In my mind these plugins not only work differently, but also serve different purposes, thus different plugin implementations.
From technical standpoint:
1. name_block has corresponding whitelist, while RR filtering plugin has not. In fact even time-restricted rules are a bit overkill for this kind of job.
2. name_block plugin will block entire response if any ANS RR matched, RR filtering plugin will remove only matched ANS RR (also it will filter EXTRA section).
3. finally, name_block does not guarantee blocking due to aliasesLimit. 

This implementation differences are coming from use-case differences between this plugins:
name_block is a tool to implement content blocking (like ads, porn, social etc), while RR filtering is a way for implementing advanced scenarios (ok, hacks), like "DNS-based routing in dual-stack networks" as i call it in my case. This two use-cases implies different models of working with blacklist file: blocking plugin usually have some king of CRON script, that will rewrite blacklist periodically, while second case is more like one-time configuration (akin of configuring forwarding rules for provider internal DNS zone). Mixing this two use-cases together will introduce complexity and breaking changes to every possible blocklist updater (like generate-domains-blacklist.py in this repo) for anyone who will decide to use both new opportunities with old (sometimes self-written) update scripts.

Saying all the above, i`m still open to discussion and changing this PR to reflect any arrangements made.

Cheers.

jedisct1:
If that turns out to be a feature that more people need, it can be implemented, but in that case, I really think that we should extend the existing syntax rather than introduce new files. That would be very confusing.

Issue #1204, opened 2020-02-25
Some specified servers don't work at all (Google, geekdns-hk...)


JoshDaone:
## Subject

* [ ] I'd like to contribute to the project (code, documentation, website, advocacy, integration, ...)
* [ ] I want to run my own DNSCrypt and/or DoH server
* [ ] I'd like to have my existing server added to a public list of encrypted DNS servers
* [ ] I'd like to announce a new application/website/article related to the project
* [x] I have questions about the DNSCrypt or Anonymized DNSCrypt protocols
* [ ] I am a MacOS developer and I'd like to contribute a GUI for dnscrypt-proxy
* [ ] I am an Android developer and I'd like to contribute an Android app
* [ ] I am a technical writer and I'd like to help write an IETF Internet Draft for the DNSCrypt protocol
* [ ] Other

## Description

Hi there, running latest version of DNSCrypt-proxy on Windows 10 1909 x64, and my case is servers like Google and geekdns-hk never gets connected from day one, no socks5/http proxy configured. My config show as below:

```
server_names = ['cloudflare', 'google', 'nextdns', 'geekdns-doh-east', 'geekdns-hk', 'rubyfish-ea', 'rubyfish-uw'] 
//defined Google and geekdns-hk
...
# Use servers reachable over IPv4
ipv4_servers = true

# Use servers reachable over IPv6 -- Do not enable if you don't have IPv6 connectivity
ipv6_servers = false

# Use servers implementing the DNSCrypt protocol
dnscrypt_servers = false

# Use servers implementing the DNS-over-HTTPS protocol
doh_servers = true


## Require servers defined by remote sources to satisfy specific properties

# Server must support DNS security extensions (DNSSEC)
require_dnssec = false

# Server must not log user queries (declarative)
require_nolog = true

# Server must not enforce its own blacklist (for parental control, ads blocking...)
require_nofilter = false

# Server names to avoid even if they match all criteria
disabled_server_names = ['doh.li', 'brahma-world', 'aaflalo-me-nyc', 'doh-westus-pi-dns', 'aaflalo-me', 'dns.sb', 'cz.nic']
```
Loglevel set to debug and get no Google/geekdns record at all:

```
9160 [2020-02-25 20:45:45] [NOTICE] dnscrypt-proxy 2.0.39
9161 [2020-02-25 20:45:45] [NOTICE] Network connectivity detected
[...]

PR #1207, opened 2020-02-26
A few more tiny doc fixes

welwood08:
Most of this should be uncontroversial, but I'd like to be sure my change to the `server_names` comment is still correct.

I must have noticed last time I looked closely at the relevant section of code that it didn't match my expectation for the `require_*` flags, and I just didn't realise why my expectations were the way they were until after responding to #1204.


jedisct1:
Thank you!

PR #1214, opened 2020-03-02
Don't log a server for blocked responses

Dragonfir3:


Dragonfir3:
Sorry, I only use domain blocking and the change made sense, but I checking that it also affects IP blocking, where the server works.

jedisct1:
Even name blocking may need an interaction with a server, in order to block `CNAME` pointers.

Dragonfir3:
Only need interaction with server if checks domain alias, right?
If so, I could filter it by adding a boolean (I thinking PluginsState.noServed)

welwood08:
Is it necessary to always pick a server at the start of every incoming query even if it's not used? I feel like I'd solve this by making the server selection lazy and then the logging plugin doesn't need a special flag. There could even be a minor performance benefit if all the locking and sorting of the server list can be avoided for some requests. All 3 references to `proxy.processIncomingQuery(...)` call `proxy.serversInfo.getOne()` in-line, so it smells like something to refactor anyway.

jedisct1:
Indeed, `getOne()` can be called only when necessary, so that it will also be copied to the plugin state when needed.

Dragonfir3:
I made the changes to filter only direct rejects, I'm testing it and log seems correct now.
In any case, the proposal of @welwood08 is more efficient, only call to server selector if this is necessary.

jedisct1:
> I made the changes to filter only direct rejects, I'm testing it and log seems correct now

Mmm... doesn't it also remove the server if the reject was due to a `CNAME`?

Dragonfir3:
No, because only removes it with reject without alias. 

```
plugin_block_name.go

	if aliasFor != nil {
		reason = reason + " (alias for [" + *aliasFor + "])"
	} else {
		pluginsState.noServed = true
	}
```

Also, I moved this if block several lines below, when the reject is confirmed.

query.log:
```
[2020-03-04 10:49:34]   192.168.2.22    incoming.telemetry.mozilla.org  A       REJECT  0ms     -
[2020-03-04 10:49:42]   192.168.2.22    g.live.com      A       REJECT  37ms    dnscrypt.eu-nl
```

welwood08:
I'm guessing since you deleted the branch this PR was based on, you plan to submit a new PR with different commits once you're done testing and that will make things clearer.

Regarding my proposal, since there's no objections I'd be glad to look into how feasible it is to pick server lazily. I don't expect it to be done before the quick logging change though, so by all means continue.

Dragonfir3:
OK. New PR opened.

PR #1259, opened 2020-03-30
Respect proxy.mainProto in forward plugin

komapa:
We have a problem very similar to the one described in http://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2008q2/002103.html
We cannot use UDP for the upstream servers and we were planning on using dnsmasq but as described later in that thread, dnsmasq would send UDP clients as UDP to the upstreams and there is no way to force it to do TCP to the upstreams when the client uses UDP.

We can achieve similar functionality with `dnscrypt-proxy` with a config like the following:

```
      listen_addresses = ['127.0.0.1:53']
      offline_mode = true
      force_tcp = true
      ipv4_servers = true
      ipv6_servers = false
      block_ipv6 = true
      timeout = 5000
      keepalive = 30
      cache = false
      forwarding_rules = 'forwarding-rules.txt'
      log_file = '/dev/stdout'
      use_syslog = false
      log_level=0

      [query_log]
      file = '/dev/stdout'
      format = 'tsv'
      ignored_qtypes = ['DNSKEY', 'NS']
```

The current problem as it exist in `dnscrypt-proxy` right now is that the forwarding plugin does not respect the `force_tcp = true` option. This PR addresses that problem.

I was thinking to change the format of `forwarding-rules.txt` to be `domain IPs proto` instead of just `domain IPs`. I think that would work equally well but seems like it might be overkill if we already have the `force_tcp` option.

Please let me know what you think and I can add unit tests and stuff like this if possible.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -443,7 +443,7 @@ func (proxy *Proxy) processIncomingQuery(clientProto string, serverProto string,
 	if len(query) < MinDNSPacketSize {
 		return
 	}
-	pluginsState := NewPluginsState(proxy, clientProto, clientAddr, start)
+	pluginsState := NewPluginsState(proxy, clientProto, clientAddr, serverProto, start)
 	serverName := "-"
 	needsEDNS0Padding := false
 	serverInfo := proxy.serversInfo.getOne()
@@ -543,12 +543,6 @@ func (proxy *Proxy) processIncomingQuery(clientProto string, serverProto string,
 			}
 			if response == nil {
 				response = serverResponse
-			}
-			if err != nil {
-				pluginsState.returnCode = PluginsReturnCodeNetworkError
-				pluginsState.ApplyLoggingPlugins(&proxy.pluginsGlobals)
-				serverInfo.noticeFailure(proxy)
-				return
 			}
 			if len(response) >= MinDNSPacketSize {
 				SetTransactionID(response, tid)
--- a/dnscrypt-proxy/plugin_forward.go
+++ b/dnscrypt-proxy/plugin_forward.go
@@ -14,6 +14,7 @@ import (
 type PluginForwardEntry struct {
 	domain  string
 	servers []string
+	proto   string
 }
 
 type PluginForward struct {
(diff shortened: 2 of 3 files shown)


jedisct1:
This is weird. Why would you need TCP to talk to your router?

I'm also not sure that captive portals support TCP at all.

komapa:
@jedisct1 let me know if the new information makes sense or you are still not sure about the usefulness of the change.

jedisct1:
Thanks for the context!

Let's use `force_tcp`, then. If it turns out to be an issue, we can always make the configuration finer grained later.

Instead of having a copy of the protocol (which is constant) in each entry, maybe the protocol could be in `pluginsState` instead?

dnscrypt-proxy/proxy.go
```diff
@@ -544,12 +544,6 @@ func (proxy *Proxy) processIncomingQuery(clientProto string, serverProto string,
 			if response == nil {
 				response = serverResponse
 			}
-			if err != nil {
-				pluginsState.returnCode = PluginsReturnCodeNetworkError
-				pluginsState.ApplyLoggingPlugins(&proxy.pluginsGlobals)
-				serverInfo.noticeFailure(proxy)
-				return
-			}
```
komapa, line 552: Unreachable code as it is duplicate of the block some lines above

komapa:
@jedisct1 I think it makes sense to be part of pluginState so I made the change. Let me know if this is what you had in mind. I destroyed my test setup and cannot verify it works as the previous solution but reading the code, it takes the value from exactly the same main place so I am pretty confident it works :)

jedisct1:
Thank you!

komapa:
Thank you for merging @jedisct1. When do you plan to release a new version so I can stop patching it locally? :) 

Thank you again!

komapa:
Bump!

jedisct1:
Honestly I don’t know. A new version cannot be released just for this, and the current one could get some CI tests for the new features before a new version is tagged.

komapa:
Yeah, it was not my idea to get a new version just for this, but more if there is some cadence on releases. I will wait patiently for the next release :)

PR #1266, opened 2020-04-04
Change URL of Block Spotify ads

spithash:
The url of the Spotify-HOSTS.txt is changed. Path of "/filter/" is now "/filters/"


jedisct1:
Thank you!

PR #1298, opened 2020-04-26
Plugins init wait

### LINKED ISSUE #1295: random panic at blacklist load
I use `master` 7d0e144 version self compiled for `Linux arm64`. Sometimes boot process exits with panic as it reaches `Loading the set of blocking rules`. At first I though that my blacklist file is corrupted due to bad sd card, but it is not. This happens sometimes when I reboot my raspberry pi 3.
[](https://postimg.cc/rK9QdvK7)

Stack trace is pointing to [here](https://github.com/DNSCrypt/dnscrypt-proxy/blob/7d0e1440e1170dc1c878101e72697f3fb84a20cf/dnscrypt-proxy/plugins.go#L265), and it turns out that pointer to slice is nil. [Gist](https://play.golang.org/p/UGYdgAqxZCY) of the problem.

vbauerster:
This PR should fix #1295.

I've removed `RWMutex` from `PluginsGlobals` as it's `Lock` method never called anyway:
[](https://postimg.cc/bZ023f1R)

### DIFF
--- a/dnscrypt-proxy/plugins.go
+++ b/dnscrypt-proxy/plugins.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"net"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/jedisct1/dlog"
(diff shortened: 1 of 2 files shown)


jedisct1:
Pointers to plugins are used so that they can eventually be reloaded without blocking or restarting the whole server. 

The feature hasn't been implemented, but that's the intent (this is also why plugins have a `Reload` method).

jedisct1:
This is also why the lock is here. Even if pointers can be atomically exchanged (is it true in Go?), we still have to lock the set so that plugins can be removed or added.

PR #1302, opened 2020-04-29
Update deps

milgradesec:
updates golang.org/x/crypto
updates golang.org/x/sys
updates gopkg.in/yaml.v2


jedisct1:
There doesn't seem to be any changes here, except `Getrlimit`/`Setrlimit` changes for `aarch64`, but `dnscrypt-proxy` doesn't use these functions anywhere.

PR #1325, opened 2020-05-20
Update EnergizedProtection URLs

spithash:
EnergizedProtection url links have changed, it seems they had to delete them from github and moved them to their self hosted domain (block.energized.pro).


jedisct1:
Thank you!

PR #1327, opened 2020-05-21
Added Block spying and tracking on Windows

spithash:
Reduces a great amount of telemetry on Microsoft based operating systems, for those who need it :)


jedisct1:
Thanks!

PR #1332, opened 2020-05-27
Fallback to cache_file avoiding termination for not offline_mode

lifenjoiner:
Retry after MinimumPrefetchInterval.
To be more robust to startup on a poor network.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -712,8 +712,11 @@ func (config *Config) loadSource(proxy *Proxy, requiredProps stamps.ServerInform
 	}
 	source, err := NewSource(cfgSourceName, proxy.xTransport, cfgSource.URLs, cfgSource.MinisignKeyStr, cfgSource.CacheFile, cfgSource.FormatStr, time.Duration(cfgSource.RefreshDelay)*time.Hour)
 	if err != nil {
-		dlog.Criticalf("Unable to retrieve source [%s]: [%s]", cfgSourceName, err)
-		return err
+		if len(source.in) <= 0 {
+			dlog.Criticalf("Unable to retrieve source [%s]: [%s]", cfgSourceName, err)
+			return err
+		}
+		dlog.Infof("Downloading [%s] failed: %v, use cache file to startup", source.name, err)
 	}
 	proxy.sources = append(proxy.sources, source)
 	registeredServers, err := source.Parse(cfgSource.Prefix)
--- a/dnscrypt-proxy/sources.go
+++ b/dnscrypt-proxy/sources.go
@@ -135,7 +135,8 @@ func fetchFromURL(xTransport *XTransport, u *url.URL) (bin []byte, err error) {
 }
 
 func (source *Source) fetchWithCache(xTransport *XTransport, now time.Time) (delay time.Duration, err error) {
-	if delay, err = source.fetchFromCache(now); err != nil {
+	var errCached error
+	if delay, errCached = source.fetchFromCache(now); errCached != nil {
 		if len(source.urls) == 0 {
 			dlog.Errorf("Source [%s] cache file [%s] not present and no valid URL", source.name, source.cacheFile)
 			return
@@ -171,6 +172,10 @@ func (source *Source) fetchWithCache(xTransport *XTransport, now time.Time) (del
 		dlog.Debugf("Source [%s] failed signature check using URL [%s]", source.name, srcURL)
 	}
 	if err != nil {
+		if errCached == nil {
+			dlog.Debugf("Source [%s] failed to update, retry after %v", source.name, delay)
+			return delay, nil // fallback avoids termination, sleep MinimumPrefetchInterval
+		}
 		return
 	}
 	source.writeToCache(bin, sig, now)
--- a/dnscrypt-proxy/sources.go
+++ b/dnscrypt-proxy/sources.go
@@ -135,8 +135,7 @@ func fetchFromURL(xTransport *XTransport, u *url.URL) (bin []byte, err error) {
 }
 
 func (source *Source) fetchWithCache(xTransport *XTransport, now time.Time) (delay time.Duration, err error) {
-	var errCached error
-	if delay, errCached = source.fetchFromCache(now); errCached != nil {
+	if delay, err = source.fetchFromCache(now); err != nil {
 		if len(source.urls) == 0 {
 			dlog.Errorf("Source [%s] cache file [%s] not present and no valid URL", source.name, source.cacheFile)
 			return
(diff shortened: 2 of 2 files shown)


welwood08:
My first concern here is with failing unit tests. Not sure why tests aren't running on pull requests (maybe they were lost in the migration to GitHub Actions), but I expect merging this as-is would result in failed builds.

It looks like there are 2 underlying causes for the failures:
- In `cache_expired` tests, download errors are expected to be returned but should no longer be expected. Since this is the behaviour you're changing, this makes sense and the test should be updated. My first instinct to resolve this only results in more failures, so I'll let you have a look before I try again.
- In `download_no-urls` and `download_path-err` tests, cache errors are unexpectedly `nil`. Fixing this requires 1 return statement to explicitly use the renamed error variable (instead of the current implicit return).

dnscrypt-proxy/sources.go
```diff
@@ -135,7 +135,8 @@ func fetchFromURL(xTransport *XTransport, u *url.URL) (bin []byte, err error) {
 }
 
 func (source *Source) fetchWithCache(xTransport *XTransport, now time.Time) (delay time.Duration, err error) {
-	if delay, err = source.fetchFromCache(now); err != nil {
+	var errCached error
+	if delay, errCached = source.fetchFromCache(now); errCached != nil {
 		if len(source.urls) == 0 {
 			dlog.Errorf("Source [%s] cache file [%s] not present and no valid URL", source.name, source.cacheFile)
 			return
```
welwood08: This return needs to reference the new `errCached` variable explicitly otherwise it always returns a nil `err`. Should fix unit test failures for `download_no-urls` and `download_path-err`.

lifenjoiner: I have updated the code, please review.

1. cache expired or read cache failed: return delay 0
2. pass `errCached` forward for empty urls

lifenjoiner:
I think here it is expected that:
1. Only startup update (`NewSource`) failing falls back to use even expired cache file;
2. Update daemon thread (callee `PrefetchSources`) failing falls back to a shorter `MinimumPrefetchInterval` sleep.

dnscrypt-proxy/sources.go
welwood08: The `download_no-urls` and `download_path-err` test failures are not yet resolved, the return statement on this line is still always returning `nil` error even when there is a fatal error.

welwood08: For safety, since there are other returns before you set `err`, I would revert the`if` statement and then set `errCached = err` afterwards.

welwood08:
For what it's worth, test failures have increased since I last checked. I'm not comfortable with hiding errors under certain conditions and overriding `delay` to indicate that has happened to some callers.

After considering your goal more carefully, I think I can propose an alternate solution that causes no unit test failures (because the real changing behaviour is outside of unit test coverage, which isn't your problem).

If you notice the `in` field of the `Source` struct is only set after either successfully reading from or writing to cache, then the caller of `NewSource` can handle the returned error non-fatally if that field is non-nil.

I have a working prototype of this proposal but I don't want to step on your toes since it's your idea, so I'll just show this for now:
```
[will@brahms dnscrypt-proxy]$ git diff --stat
 dnscrypt-proxy/config.go | 7 +++++--
 1 file changed, 5 insertions(+), 2 deletions(-)
[will@brahms dnscrypt-proxy]$ go test ./...
ok  	github.com/DNSCrypt/dnscrypt-proxy/dnscrypt-proxy	0.341s
```

I'm happy to review whichever path you choose, but I couldn't see an easy way to make the tests work with your current solution.

[EDIT: simplified proposal]

lifenjoiner:
@welwood08 

Thank you very much for your suggestions and review.

Considering to pass the tests, I changed my thought and reorganized the code:
1. Keep all the errors to be returned untouched;
2. Increase the delay to `prefetchDelay` for 'good' cache (maybe expired) but empty source urls;
3. From higher level (caller), ingnore the downloading error from `NewSource` when startup (cache loaded).

I run the test `go test -mod vendor`, on Win10 1909, go version from 1.14.3 down to 1.13, but unfortunately it always gives `checks:  1981 passed  0 todo  28 failed       (total)` before and after my patch :(
All the unexpected errors are `invalid port` that say error string not match that like this:
```
sources_test.go:371: Unexpected error
    Checker:  Match
    Regex:    (string) (len=12) 'invalid port'
    Actual:   (string) (len=41) 'lookup 127.0.0.1:6327900000: no such host'
```

Please review again.

lifenjoiner:
Removed an unnecessary blank.

lifenjoiner:
~~2. Increase the delay to prefetchDelay for 'good' cache (maybe expired) but empty source urls;~~

And I thing I should minimize my patch, and keep the delay to use `MinimumPrefetchInterval`.

dnscrypt-proxy/config.go
```diff
@@ -712,8 +712,12 @@ func (config *Config) loadSource(proxy *Proxy, requiredProps stamps.ServerInform
...
 	source, err := NewSource(cfgSourceName, proxy.xTransport, cfgSource.URLs, cfgSource.MinisignKeyStr, cfgSource.CacheFile, cfgSource.FormatStr, time.Duration(cfgSource.RefreshDelay)*time.Hour)
 	if err != nil {
-		dlog.Criticalf("Unable to retrieve source [%s]: [%s]", cfgSourceName, err)
-		return err
+		if len(source.in) > 0 { // cache loaded, ignore download failure
+			dlog.Infof("Downloading [%s] failed: %v, use cache file to startup", source.name, err)
+		} else {
+			dlog.Criticalf("Unable to retrieve source [%s]: [%s]", cfgSourceName, err)
+			return err
+		}
```
welwood08, line 718: If you swap these 2 around so the `if` checks the fatal condition, then the non-fatal path doesn't need to be in an `else` block. This would also match the style used later in the file to check the next "maybe-fatal" condition.

welwood08 (commented):
Looks good to me other than 1 small nit. I've tested on linux to confirm expected behaviour. Don't worry about the unit tests failing on Windows, that's not currently part of the automated tests and is something I'll look into at some point.

dnscrypt-proxy/config.go
lifenjoiner: Yeah, you are right. It gives a tighter code and matches the code style.
I took your advice, and appreciate your patience to guide me, even though you got the point quicker. This really helps beginners who's willing to contribute to get involved. Thanks!

welwood08 (approved):
Thanks for the contribution!

jedisct1:
Thank you so much to both of you! This is a nice addition!

PR #1334, opened 2020-05-27
Update link to Spam404 rule

IceCodeNew:
Though the old link is still available, the rule file has stop updating for a long while.


jedisct1:
Thank you!

PR #1335, opened 2020-05-27
Added Block all known NSA / GCHQ / C.I.A. / F.B.I. spying servers

IceCodeNew:
For those who need it :-P


jedisct1:
Mmm... that list has too many false positives to be used for anything.

```text
0.0.0.0 dns.ns1.edgecastcdn.net
0.0.0.0 dns.ns2.edgecastcdn.net
0.0.0.0 dns.ovh.net
0.0.0.0 huey.disney.com
0.0.0.0 dns.princeton.edu
```

jedisct1:
I'd rather avoid that list.

IPs used by government agencies are unlikely to be public and well-known.

The list blocks entire legitimate ISPs (Orange, Telia, Rackspace, OVH, Chello...) so it would just break people's Internet while providing little value.

IceCodeNew:
You are right.

PR #1341, opened 2020-05-29
Minor update to GH Actions workflow

welwood08:
Not sure if it's intended that CI didn't run on pull requests, but I believe it's safe to do so since GitHub explicitly won't expose secrets to PRs from a fork.

I've used my earlier experiment as reference to filter the paths that will trigger CI for push and pull request events.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -1,4 +1,21 @@
-on: push
+on:
+  push:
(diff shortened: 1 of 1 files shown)


jedisct1:
Having it run on pull request is definitely useful. I thought it was already the case, but apparently not. Thanks for fixing this!

PR #1350, opened 2020-06-03
Add comments for domains-blacklist.conf, with a little sorting on it.

IceCodeNew:
0. Add more comments so it should be much easier for anyone to get understanding how to choose the rules which is delivered in varies levels.
1. Sort rules from Energized so it is ordered in the sort of size, which would make sense.


IceCodeNew:
2. Add rule from AdAway.

jedisct1:
Excellent, thank you!

PR #1353, opened 2020-06-09
update domains-blacklist.conf

Huhni:
When I checked out the new **SANS Institute suspicious domains** lists I noticed that they are not available anymore, which led me to also look at some other sources.

It turns out that many of the lists enabled by default are already included in the Energized Blu list (that is also enabled by default).

Also some of the lists consist almost exclusively of rules for removing elements or paths like `/ads`, but barely contain any domains, therefore I also removed them.

I'm also not sure if easylist china should be enabled by default, but I don't know the user base.

This is unrelated to the pull request, but I found a big collection of filter lists here https://firebog.net/ which might be useful


jedisct1:
Sad for SANS' lists, but if they were not maintained any more, removing them was the right thing to do.

Thanks a lot for these updates and changes!

PR #1358, opened 2020-06-10
sources_test: set bit 16 of the port instead of adding zeros

jedisct1:
An attempt at fixing the tests on Windows 10 with 32 bit builds.

/cc @welwood08 @lifenjoiner


welwood08:
Technically you're setting bit 16, but either way I've tested GOARCH=386 on linux and tests now pass.
I'll leave it to you to merge in case you want to fix the commit message.

lifenjoiner:
Finally, we can make it better. Thank you all!

PR #1363, opened 2020-06-14
Return earlier from unreachable host to download sources

lifenjoiner:
Reduce coding running, although it is cached `cache_neg_max_ttl` seconds.

### DIFF
--- a/dnscrypt-proxy/xtransport.go
+++ b/dnscrypt-proxy/xtransport.go
@@ -280,16 +280,16 @@ func (xTransport *XTransport) resolveUsingResolvers(proto, host string, resolver
 }
 
 // If a name is not present in the cache, resolve the name and update the cache
-func (xTransport *XTransport) resolveAndUpdateCache(host string) error {
+func (xTransport *XTransport) resolveAndUpdateCache(host string) (net.IP, error) {
 	if xTransport.proxyDialer != nil || xTransport.httpProxyFunction != nil {
-		return nil
+		return nil, nil
 	}
 	if ParseIP(host) != nil {
-		return nil
+		return nil, nil
 	}
 	cachedIP, expired := xTransport.loadCachedIP(host)
 	if cachedIP != nil && !expired {
-		return nil
+		return nil, nil
 	}
 	var foundIP net.IP
 	var ttl time.Duration
@@ -327,12 +327,12 @@ func (xTransport *XTransport) resolveAndUpdateCache(host string) error {
 			foundIP = cachedIP
 			ttl = ExpiredCachedIPGraceTTL
 		} else {
-			return err
+			return foundIP, err
 		}
 	}
 	xTransport.saveCachedIP(host, foundIP, ttl)
 	dlog.Debugf("[%s] IP address [%s] added to the cache, valid for %v", host, foundIP, ttl)
-	return nil
+	return foundIP, nil
 }
 
 func (xTransport *XTransport) Fetch(method string, url *url.URL, accept string, contentType string, body *[]byte, timeout time.Duration) ([]byte, *tls.ConnectionState, time.Duration, error) {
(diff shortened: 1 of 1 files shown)


lifenjoiner:
Background:
```
    --- FAIL: TestNewSource/cache_expired,_download_open-err/7-1 (0.21s)
        sources_test.go:371: Unexpected error
            Checker:  Match
            Regex:    (string) (len=12) 'invalid port'
            Actual:   (string) (len=116) 'Get "http://127.0.0.1:6202600000/8/empty.md": dial tcp: address 127.0.0.1:6202600000:443: too many colons in address'
```
After some investigation for my previous test error message (`v2.0.43`, before commit https://github.com/DNSCrypt/dnscrypt-proxy/commit/f4d519092b4050a42fa3ec6060461dc1ac47f3eb), I figured out how the `:443` was appended:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/f9268be0c21b2deda6a2375a7b2c809a4ebf1be6/dnscrypt-proxy/xtransport.go#L135-L156
It is the `stamps.DefaultPort` and function `rebuildTransport`.

And this only related to the `sources` file downloading:
ExtractHostAndPort

Now, on `v2.0.44`, it won't form the too big port to trigger this error on 32-bit test.

I supposed a bad case:
The `sources.url` is not correctly configured. Maybe it is non-existed (typos), no longer available, or blocked ...
There is not much possibility to happen, but I made this patch in case.

BTW:
`CI` failed with `raw.githubusercontent.com` and `download.dnscrypt.info` were returned `NXDOMAIN`. Strange ...

lifenjoiner:
`CI` is very useful. I have fixed the first commit.

lifenjoiner:
Frankly to say, it's a very small commit. Maybe it's just right in logic, as the `resolveAndUpdateCache` is designed to be. If you think it doesn't worth it, you can close it. I'll still be glad to contribute to make `dnscrypt-proxy` better if there is a chance.

welwood08:
Apologies this wasn't looked at sooner - please don't be discouraged, it certainly looks like some excellent investigation! I'm not too familiar with the `XTransport` internals (which might explain why I wrongly attributed its behaviour to Windows 10), so maybe @jedisct1 can chime in if I miss anything.

While trying to understand why the original code does things the way it does, I noticed that at least the system resolver path can return no error but also no IPs if only IPv6 results are received and the configuration requires IPv4, or vice-versa. This would lead to the host having a nil cached IP and it's unclear how `dialer.DialContext` behaves when passed a hostname it still needs to resolve.

Unfortunately, it looks like your current solution will treat every DoH query as NXDOMAIN if a user has configured a proxy such as Tor because of the 1st `return nil, nil` in `resolveAndUpdateCache`. I don't know exactly how things should work in this case, but presumably the IP cache is not used. An improvement might be to do the check for `foundIP == nil` right before the `saveCachedIP` call and return a new error, this way there's no need to return the IP (and as a bonus it doesn't create a cache entry with a nil IP).

I suggest also that the `ExtractHostAndPort` method is worth some attention: the problem you discovered was that when it fails to parse the port as an integer it leaves the port in the returned host string and also returns the default port. Perhaps if `strconv.Atoi` fails, the bad port should be trimmed from the returned host?

jedisct1:
Hi,

And sorry for the delayed response. I'm currently traveling and have limited internet connectivity. Reading issues is easy, but reviewing code, not so much :)

This part of the code is quite tricky. Being able to bootstrap everything without a valid DNS configuration, Onion services support, IPv4/IPv6 restrictions, IPs that expire but that still need to be reused if the configuration is unusable... all these things make it complicated and fragile. 

Changing one small thing may break something else, and we don't have tests covering all the cases.

I'll try to take a look over the weekend.

lifenjoiner:
@welwood08 
You're right, the previous issue is because of 'bad' port isn't thrown out.
And you mentioned something like 'IPv6' and the `DoH` by proxy, I missed. I thought I considered all ...

So, for this multi-functional and well-developed program, that's true `Changing one small thing may break something else`! @jedisct1 may have a better solution/decision with a bigger picture.

This patch is intended to short the return time from a bad case. But if it costs too much, leave it and we can treat it as a normal one.

And one more situation: `dnscrypt-proxy` is an up-stream server of a local DNS. If it download the `sources` file, it would still go through the system configuration (DNS) ... And `resolveAndUpdateCache` is a little tricky, downloading may not use the answer here. Seems there is no direct effect ...

I will rethink about it.

I learnt a lot from talking to you.

PR #1365, opened 2020-06-16
Disable Sysctl list and AdAway, introduce GameIndustry.eu; Remove Chinese IT companies' top domain from whitelist.

IceCodeNew:
* The host file from http://sysctl.org/cameleon is no longer updated, therefore it should be disabled.

* Introduce a new rule maintained by GameIndustry.eu. I only pick the rule sets that NextDNS provides to its customers of their choice, as these rule sets are generally seen as stable and reliable.
However I don't play game so much, there is no way to perform a fully test on my side. There is no FP detected during the couple of days while I using this rule set. And I've gone through the entire contents of the host file in roughly, the entries all seem reasonable to me.

~~It doesn't take long for jedisct1 add baidu.com, and 163.com into whitelist after I introduced AdAway's ruleset into the configuration file, so I guess that the AdAway rule set must have presented a lot of false positives.
However, these Chinese IT companies are notorious for their extensive user-tracking tactics. Whitelist their top domain may not a good idea.
My suggestion is to simply disable the ruleset present FP, and let software like ABP or AdGuard do the most elaborate work. - Blocking on the DNS level has its limitations.~~

### DIFF
--- a/utils/generate-domains-blacklists/domains-blacklist.conf
+++ b/utils/generate-domains-blacklists/domains-blacklist.conf
@@ -27,7 +27,7 @@
 file:domains-blacklist-local-additions.txt
 
 # AdAway is an open source ad blocker for Android using the hosts file.
-https://raw.githubusercontent.com/AdAway/adaway.github.io/master/hosts.txt
+# https://raw.githubusercontent.com/AdAway/adaway.github.io/master/hosts.txt
 
 # Malware domains
 https://mirror1.malwaredomains.com/files/justdomains
--- a/utils/generate-domains-blacklists/domains-blacklist.conf
+++ b/utils/generate-domains-blacklists/domains-blacklist.conf
@@ -60,7 +60,7 @@ https://pgl.yoyo.org/adservers/serverlist.php?hostformat=nohtml
 # https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt
 
 # Sysctl list (ads)
-http://sysctl.org/cameleon/hosts
+# http://sysctl.org/cameleon/hosts
 
 # KAD host file (fraud/adware) without controversies
 # https://raw.githubusercontent.com/PolishFiltersTeam/KADhosts/master/KADhosts_without_controversies.txt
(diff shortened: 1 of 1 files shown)


IceCodeNew:
ruleset `GameIndustry.eu` contains entry `0.0.0.0 t.co`, which may cause trouble. Disable this ruleset by default.

utils/generate-domains-blacklists/domains-blacklist.conf
```diff
@@ -161,3 +161,6 @@ https://dbl.oisd.nl/
 
 # Block spying and tracking on Windows 
 # https://raw.githubusercontent.com/crazy-max/WindowsSpyBlocker/master/data/dnscrypt/spy.txt
+
+# GameIndustry.eu - Blockieren Sie Spyware, Advertising, Analytics, dem Tracking in Spielen und in dazugehörigen Clients.
```
jedisct1: Does the description need to be in German? :)

jedisct1:
I'm not sure about removing Microsoft, Google, Baidu, etc. from the whitelist.

People may not want to block them, and false positives happen. The reason Mirosoft domains were added is that I recently updated a Windows VM and saw Edge reporting no connection due to Microsoft having been blocked by a public list. 

The whitelist can be edited by people who explicitly want to block these, but I think most people still expect Microsoft, Google, Baidu, etc. to not report a "this domain doesn't exist" response.

What we may need is proper handling of `=` prefixed whitelisted entries, so that we can e.g. whitelist `microsoft.com` and still block some longer hostnames within the zone.

But by default, for most people, not blocking a few things may be better than blocking popular domain names. At least that would be the least surprising tradeoff.

utils/generate-domains-blacklists/domains-blacklist.conf
IceCodeNew: I'll use DeepL to translate it into English.

IceCodeNew:
> What we may need is proper handling of = prefixed whitelisted entries, so that we can e.g. whitelist microsoft.com and still block some longer hostnames within the zone.

Agree. I will be looking forward to this function enhancement.

> But by default, for most people, not blocking a few things may be better than blocking popular domain names. At least that would be the least surprising tradeoff.

You are right, I just closed another PR and rebased this PR to discard all the changes I made to the whitelist.

PR #1370, opened 2020-06-18
Add note on how to enable debug logging

dne:
I just spent some time trying to understand why `log_level=0` didn't get me any debug output in the log.


jedisct1:
Oh no! This is completely intentional and the reason this environment variable is required.

Debug logs are really meant to be used only by developers. They provide information about the code, not really anything that could be relevant for actual usage, even for troubleshooting.

dne:
OK, fair enough, but then why mention level 0 at all in the example config?

In this case, I was trying to figure out why dnscrypt-proxy was silently (at log levels 1-2) filtering out some of my configured servers – I suspect that they were temporarily unresponsive. Unfortunately, when I had finally found out about `$DEBUG`, the problem had gone away.

jedisct1:
Placebo effect 😀

PR #1372, opened 2020-06-18
Add shiftleft scan

prabhu:
Hi,

I noticed that you had tried shiftleft scan in the past. Scan is reporting a number of vulnerabilities and a handful of license conditions for this repo. Please have a look.

https://github.com/prabhu/dnscrypt-proxy/runs/785992095?check_suite_focus=true#step:4:71


jedisct1:
I tried it on different projects, but the `upload-sarif` action failed 🤷

jedisct1:
Let's give it a new try!

prabhu:
upload-sarif is a github official step that works only on repos enabled for code scanning. Once the feature is enabled it should work.

PR #1375, opened 2020-06-19
ConfigFile change to allowlist and blocklist

ianbashford:
I've made a few changes to move to 'allowlist' and 'blocklist' naming for the config files.
I've chosen an approach where the old tags are still supported -- it felt too significant to erase the old names overnight...  
In the interest of small compact PRs, I've just put enough in here to respect the external interfaces (filenames etc.).   
If this change is something you approve of, then I can continue to refactor some of the golang class/file names too...


ianbashford:
I think I'm getting the same failures as after your last commit.

jedisct1:
Hi Ian!

And thanks for this!

Maybe we can avoid "list" (`allowlist` doesn't sound natural and `blocklist` looks like a typo) and be more specific?

`[ip_blacklist]` => `[blocked_ips]`
`[blacklist]` => `[blocked_names]`
`[whitelist]` => `[allowed_names]` or even `[always_allowed_names]`?

jedisct1:
Also maybe a warning can be printed when the legacy version is used, in order to encourage people  to upgrade their configuration, so that we can eventually remove the compatibility layers.

ianbashford:
Thanks for the comments -- I struggled with the choice a little too ( clearlist / safelist I rejected as unclear or even misleading ).  
`blocked_{ips|names}` are both much clearer.
I fear `[always_allowed_names]` doesn't make clear the scheduler options that mean they're not actually always allowed.  
What do you think of being a bit clearer about what is meant by 'names' ?  e.g.:
```
[blocked_ips]
[blocked_hosts]
[allowed_hosts]
```

welwood08:
It could be argued that `hosts` might be technically less clear than `names` in this context. Host names have a specific syntax limiting them to only alphanumeric and hyphen characters (see the grammars defined in [RFC 952](https://tools.ietf.org/html/rfc952)), whereas for example it's perfectly valid to query for a SRV record with the name `_service._proto.name`. Host names are a subset of domain names. See [RFC 7719 section 2](https://tools.ietf.org/html/rfc7719#section-2) and follow the rabbit hole for the gory details, particularly [RFC 2181 section 11](https://tools.ietf.org/html/rfc2181#section-11).

ianbashford:
thanks @welwood08  - names it is.   I'll work on those changes...

dnscrypt-proxy/example-dnscrypt-proxy.toml
```diff
@@ -479,25 +479,25 @@ cache_neg_max_ttl = 600
...
 ##   fe80:abcd:*
 ##   192.168.1.4
 
-[ip_blacklist]
+[blocked_ips]
 
   ## Path to the file of blocking rules (absolute, or relative to the same directory as the config file)
 
-  # blacklist_file = 'ip-blacklist.txt'
+  # blocked_ips_file = 'blocked_ips.txt'
```
alisonatwork: Since the suggested log file name (line 500) is `blocked-ips.log` i feel like it might be a bit more consistent to also call this `blocked-ips.txt`. All the other filenames are also using kebab-case (`public-resolvers.md`, `service-install.bat` etc) so perhaps it would be best to stick with kebab-case for filenames and snake_case for the config options.

dnscrypt-proxy/example-dnscrypt-proxy.toml
```diff
@@ -444,10 +444,10 @@ cache_neg_max_ttl = 600
 
 
 ######################################################
-#        Pattern-based blocking (blacklists)        #
+#        Pattern-based blocking (blocked lists)        #
```
alisonatwork: I think it would be clearer if in the comments we consistently refer to blocklists and allowlists to describe the abstract concept, then use the "blocked" and "allowed" terminology for specific config options (`blocked_names`, `blocked_ips`, `allowed_names`). Right now the comments are a bit mixed between "blocklists" and "blocked lists", which is a bit confusing. I am not too fussed which one we choose, just as long as we are consistent.

From my side, I think "blocklist" is clearer as a noun than "blocked list", since to me "blocked list" sounds like the list itself is blocked, rather than that it is a list containing blocks.

dnscrypt-proxy/example-dnscrypt-proxy.toml
ianbashford: thanks, this does look much more consistent now - inc in latest PR update

dnscrypt-proxy/example-dnscrypt-proxy.toml
ianbashford: Agreed and updated in latest PR update

dnscrypt-proxy/example-dnscrypt-proxy.toml
alisonatwork: Changes look great, thanks!

jedisct1:
Awesome!

Would it be possible to also update the CI scripts (in the `.ci` directory)?

ianbashford:
I missed those... on it

ianbashford:
CI tests are clear with changes to the new naming scheme

jedisct1:
Awesome work, Ian!

jedisct1:
Thanks a lot Alison and Will too!

ianbashford:
thanks all for the help getting this PR up to scratch
I'll start to work on a PR for the internal interfaces...

PR #1376, opened 2020-06-19
Fix dependency error and undefined type

syphyr:
Fix a few compile errors


syphyr:
Fixed by another recent commit

PR #1388, opened 2020-06-26
Add forward slashes to example stamp for consistency

hugepants:
Seems to work with or without, but makes it consistent with the toml, the documentation and the stamp calculator.

### DIFF
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -747,4 +747,4 @@ skip_incompatible = false
 [static]
 
   # [static.'myserver']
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks!

The lack of forward slashes was intentional to show that both are supported.

For some backstory, I tried to register the scheme to the IETF: https://ietf.topicbox-scratch.com/groups/uri-review/Ta2a407095e4bcb65-M3357dee85522c0774120e2b5

It was suggested to remove the forward slashes because the information being encoded was not hierarchical. There was no "paths".

So, `dnscrypt-proxy` started to support both, with the eventual plan to remove them everywhere.

Then came anonymized DNS. And now, having these slashes totally made sense again. `sdns://<relay>/<server>` can represent how to reach a server via a relay, and that can even be extended to multiple relays.

So, the `://` was kept.

So we can indeed bring the slashes back for consistency.

Thanks!

hugepants:
Good to know, thanks!

PR #1406, opened 2020-07-07
Further block/allow updates

ianbashford:
Some more of the move to block / allow lists
updates to the example files missed first time around
Modifications to the utils blocklist generation files


welwood08:
Reading some of these changes, I can't help but notice that using "allowlist" as a verb is awkward. I have no problem with the noun form (although maybe it isn't a real word and should have a space), but I think we should find ways to rewrite sentences where the verb form is used. For consistency, we should probably do the same with the verb form of "blocklist". I'd be happy to offer suggestions if you'd like, as line comments. I think there were a few instances of the verb form that went in to the previous PR too.

ianbashford:
thanks for reviewing -- you're right I think I did that again.
Always happy for inline comments if you have some specific ideas;  I'll try to push some alternatives too - I'm sure we can meet in the middle easily enough on these.

dnscrypt-proxy/example-allowed-names.txt
```diff
@@ -1,16 +1,16 @@
...
-## Rules for name-based query whitelisting, one per line
+## Rules for name-based query allowlisting, one per line
 ##
 ## Example of valid patterns:
 ##
 ## ads.*         | matches anything with an "ads." prefix
 ## *.example.com | matches example.com and all names within that zone such as www.example.com
 ## example.com   | identical to the above
-## =example.com  | whitelists example.com but not *.example.com
+## =example.com  | allowlists example.com but not *.example.com
```
welwood08: allows

dnscrypt-proxy/example-allowed-names.txt
```diff
@@ -1,16 +1,16 @@
 
 ###########################
-#        Whitelist        #
+#        Allowlist        #
 ###########################
 
-## Rules for name-based query whitelisting, one per line
+## Rules for name-based query allowlisting, one per line
```
welwood08: Rules for allowing queries based on name?

welwood08: Or i guess keep it consistent with the blocked-names comment.

dnscrypt-proxy/example-blocked-ips.txt
```diff
@@ -1,8 +1,8 @@
 ##############################
-#        IP blacklist        #
+#        IP blocklist        #
 ##############################
 
-## Response containing blacklisted IP addresses will be blocked
+## Response containing blocklisted IP addresses will be blocked
```
welwood08: Something like rules for IP-based response blocking maybe? (Obviously "blocked IP addresses will be blocked" wouldn't work!)

utils/generate-domains-blacklists/generate-domains-blocklist.py
```diff
@@ -177,16 +177,16 @@ def whitelist_from_url(url):
...
 
-def blacklists_from_config_file(
-    file, whitelist, time_restricted_url, ignore_retrieval_failure, output_file
+def blocklists_from_config_file(
+    file, allowlist, time_restricted_url, ignore_retrieval_failure, output_file
 ):
-    blacklists = {}
-    whitelisted_names = set()
+    blocklists = {}
+    allowlisted_names = set()
```
welwood08: allowed_names?

dnscrypt-proxy/example-allowed-names.txt
ianbashford: used "allows example.com but not *.example.com"

dnscrypt-proxy/example-allowed-names.txt
ianbashford: Rules for name-based query allow lists, one per line ?

utils/generate-domains-blacklists/generate-domains-blocklist.py
ianbashford: allowed_names -- much better...

dnscrypt-proxy/example-allowed-names.txt
welwood08: Confusing because it reads as if the allowlist contain rules for allow lists. The file *is* the list.

dnscrypt-proxy/example-blocked-ips.txt
ianbashford: agreed - "Rules for IP-based response blocking"

dnscrypt-proxy/example-allowed-names.txt
ianbashford: it does -- on it....

ianbashford: used your original "allowing queries based on name"

dnscrypt-proxy/example-dnscrypt-proxy.toml
```diff
@@ -507,18 +507,18 @@ cache_neg_max_ttl = 600
...
-## If a name matches a allowlist entry, the corresponding session
+## If a name matches an allowlist entry, the corresponding session
 ## will bypass names and IP filters.
 ##
 ## Time-based rules are also supported to make some websites only accessible at specific times of the day.
 
 [allowed_names]
 
-  ## Path to the file of allowlisting rules (absolute, or relative to the same directory as the config file)
+  ## Path to the file of allow list rules (absolute, or relative to the same directory as the config file)
```
welwood08, line 521: I commented on the original commit but I don't know how notifications work for that so I'll add a note here about "allowlisted" -> "allowed" a few lines below here.

ianbashford: Sorry - it's possible I did an early push and obliterated it...  these commenting systems are good but not infallible.  Like me perhaps :-)  I've made this change too -- thanks for the careful and patient reviewing.

welwood08: No worries. :-)

utils/generate-domains-blacklists/generate-domains-blocklist.py
```diff
@@ -218,33 +218,33 @@ def blacklists_from_config_file(
...
+    allowed_names |= allowlist_from_url(allowlist)
 
-    # Process blacklists
-    for url, names in blacklists.items():
+    # Process blocklists
+    for url, names in blocklists.items():
         print("\n\n########## Blacklist from {} ##########\n".format(
             url), file=output_fd, end='\n')
-        ignored, glob_ignored, whitelisted = 0, 0, 0
+        ignored, glob_ignored, allowlisted = 0, 0, 0
```
welwood08: allowed?

welwood08:
I have 1 more minor concern, but other than that it looks good to me.

I don't personally use the python script but i note that you're changing a command-line argument that could potentially be being used by other scripts. Perhaps it would be prudent to deprecate the old argument with a warning? I don't have strong feelings either way, so I'll leave that up to you.

PR #1412, opened 2020-07-10
Update to generate-domains-blocklist.py

ianbashford:
As noted by @welwood08  I've given -w a deprecation message, rather than just pulling the plug.
Not sure what you think about this, so I'll float it out for comment - it's friendlier for sure.  

» ./generate-domains-blocklist.py -w whitelist.txt
Use of -w WHITELIST has been removed. Please use -a ALLOWLIST instead.


utils/generate-domains-blocklists/generate-domains-blocklist.py
```diff
@@ -278,6 +278,11 @@ def blocklists_from_config_file(
 )
 argp.add_argument(
     "-w",
+    "--whitelist",
+    help="Deprecated.  Please use -a or --allowlist",
```
jedisct1, line 282: There’s an extra space here :)

jedisct1:
What would be ideal is to support `-w` (with the deprecation notice you added) without having it listed in the help message.

But this is probably difficult to do, if possible at all without rewriting the whole command-line parsing code.

ianbashford:
> What would be ideal is to support `-w` (with the deprecation notice you added) without having it listed in the help message. 
> But this is probably difficult to do, if possible at all without rewriting the whole command-line parsing code.

I actually took a look at that b/c I wanted to do the same thing - I only stopped b/c it wasn't 'easy'.  Let me take another look and see if I can do anything neat

utils/generate-domains-blocklists/generate-domains-blocklist.py
```diff
@@ -278,6 +278,11 @@ def blocklists_from_config_file(
 )
 argp.add_argument(
     "-w",
+    "--whitelist",
```
alisonatwork, line 281: I don't think we really need to add a new long option for an option which is deprecated.

alisonatwork: Ah wait, I see, this is a follow up to #1406 👍

utils/generate-domains-blocklists/generate-domains-blocklist.py
```diff
@@ -304,6 +309,11 @@ def blocklists_from_config_file(
 
 args = argp.parse_args()
 
+whitelist = args.whitelist
+if whitelist:
+    print('Use of -w WHITELIST has been removed. Please use -a ALLOWLIST instead.')
```
alisonatwork, line 314: I don't think you need to put `WHITELIST` here, since in both cases the type of list the user is linking is the same. I think it would be fine to just say `The option to provide a set of names to exclude from the blocklist has been changed from -w to -a.` - this way it is consistent with the help text.

Also printing the help text before you exit (`argp.print_help()`) might be friendly too.

alisonatwork:
Hi @ianbashford what is the status of this PR? Are you still going to make some changes?

jedisct1:
Yep, if you think it looks fine as-is, let's get it merged :)

alisonatwork:
I think it's fine as-is, although it would be nice to add a mini commit to just update some of the user-friendliness comments. If Ian isn't around we can just merge it now and add another small PR to tweak that stuff.

ianbashford:
Sorry -- I've been swamped by work recently.
I'll try to find some time to tidy it up -- if you're happy to merge now I'll do a mini-pr later; that'll save me being a blocker...

Issue #1413, opened 2020-07-11
Should the next version be 2.1.0?


jedisct1:
This is just a cosmetic question. After 44 versions `dnscrypt-proxy` has changed quite a lot since the initial release.

Besides bug fixes and performance improvements, tons of features have been introduced, and current versions look like a very different app compared to version 2.0.0.

The next release is going to encourage, or require minor configuration changes (black/white renaming, move the v3 list URLs). It may also be a good opportunity to remove previous backward compatibility layers for configuration options that have long been deprecated or renamed. Such changes are more expected after a major version bump rather than in a version that feels like yet another incremental release.

On the other hand, there won't be any significant new features or changes as one may expect from a major version change.

What do you people think?

@alisonatwork @ianbashford @welwood08 @hugepants @lifenjoiner @IceCodeNew @kev-null

welwood08:
I'd say every release that might cause running configurations to stop working and require manual intervention should increment a more significant version part than usual, to set appropriate user expectations when updating.

This probably should have been considered more carefully with the renaming of `tls_client_auth` to `doh_client_x509_auth` for example. I do agree with using a more significant version bump as an opportunity to clean up all previous deprecations and transition layers, otherwise there's a risk they remain around forever.

While I'd normally combine this sort of version increment with a headline feature of some kind, version numbers are plentiful and if there are external reasons to prioritise a release then a headline feature need not be a requirement.

alisonatwork:
I have a feeling that a lot of people are keeping `dnscrypt-proxy` up-to-date with a package manager, so they don't really care about the version number, they just trust that the upgrade isn't going to break their existing config.

I think it makes sense to do a larger version bump if we remove old config options, since that is at least somewhat of a hint to the package maintainers that they might need to do a bit of extra work to make the upgrade clean. We could help them out by providing a script to read in an old config and write out a new one, although i guess we'd still be dependent on them to add that to their package's post-install step... and that's assuming they're even comfortable automatically rewriting their users' configs. Another option would be for `dnscrypt-proxy` itself to automatically upgrade the config file, but that feels like adding complexity that really should belong in the package manager.

Perhaps it's worth releasing one more backward compatible version first that includes warning logs for all the deprecated options, then remove all the old options in 2.1.0 and improve the behavior of the unknown config handling.

[...]

PR #1426, opened 2020-07-27
initialize the log file before reporting config errors

alisonatwork:
Further to the discussion in #1413 here is change to the way we load the config file to first initialize the log file before reporting any errors in the config. This will hopefully make it more obvious to people who are running in an automated setup (i.e. Windows or systemd service) why something went wrong after an upgrade to an incompatible version, since normal output to stderr might otherwise get lost.

Here is the output of a manual test to show how it works:

```bash
$ ./dnscrypt-proxy.exe -config abc
[2020-07-26 20:48:35] [FATAL] Unable to load the configuration file [abc] -- Maybe use the -config command-line switch?
$ cat > abc <<EOF
log_file = "def"
unknown = "bad_config"
EOF
$ ./dnscrypt-proxy.exe -config abc
$ echo $?
255
$ cat def
[2020-07-26 20:48:48] [NOTICE] dnscrypt-proxy 2.0.44
[2020-07-26 20:48:48] [FATAL] Unsupported key in configuration file: [unknown]
```

Note that all errors returned during config will result in a fatal log due to code inside main.go (see https://github.com/alisonatwork/dnscrypt-proxy/blob/master/dnscrypt-proxy/main.go#L127) so even very aggressive log levels should still log this kind of failure to the log file or syslog.

To try make the behavior of all config errors resulting in a fatal log more clear to developers, I added a second commit which makes all of the fatal logs just return an error. There is an alternative option, which would be to replace all the "return error" lines with a fatal log. Personally I prefer returning an error, but I don't care too much either way, the main point here is to make all of them consistent.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -294,7 +294,7 @@ func findConfigFile(configFile *string) (string, error) {
 func ConfigLoad(proxy *Proxy, flags *ConfigFlags) error {
 	foundConfigFile, err := findConfigFile(flags.ConfigFile)
 	if err != nil {
-		dlog.Fatalf("Unable to load the configuration file [%s] -- Maybe use the -config command-line switch?", *flags.ConfigFile)
+		return fmt.Errorf("Unable to load the configuration file [%s] -- Maybe use the -config command-line switch?", *flags.ConfigFile)
 	}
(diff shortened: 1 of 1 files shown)


jedisct1:
This is a nice improvement!

Thank you Alison!

PR #1460, opened 2020-08-28
Building From Source

JasonCoombs:
go build


jedisct1:
Maybe that should rather be a Wiki page, that also explains how to install Go on various operating systems and distributions, the minimum required version, how to check out and compile a release instead of test code and how to notarize the result on macOS.

Dependencies are vendored, so `go build -mod vendor` can be used to allow offline installation.

PR #1486, opened 2020-10-04
user-friendly comments - follow up to #1412

ianbashford:
Apologies it was such a long wait for a couple of small tweaks.

It turned out to be quite easy to suppress the -w so it doesn't show in the help statement.  I've tweaked the message as suggested as well.
This would leave it possible to run with -w  and not exit, as it does now (line 327).  Although I suspect it might be a bit late for that now... it'd be an easy update if you like that option though.


jedisct1:
Oh, awesome, thank you!

PR #1489, opened 2020-10-07
Update example-dnscrypt-proxy.toml

lifenjoiner:
`lb_strategy` option 'pn' has been implemented, but not documented.

### DIFF
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -134,7 +134,7 @@ keepalive = 30
 
 
 ## Load-balancing strategy: 'p2' (default), 'ph', 'pn', 'first' or 'random'
-## Use the 2, half, n, 1 or random number of all by latency.
+## Randomly choose 1 of the fastest 2, half, n, 1 or all live servers by latency.
 ## The response quality still depends on the server itself.
 
 # lb_strategy = 'p2'
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -133,7 +133,7 @@ keepalive = 30
 # blocked_query_response = 'refused'
 
 
-## Load-balancing strategy: 'p2' (default), 'ph', 'pn', 'first' or 'random'
+## Load-balancing strategy: 'p2' (default), 'ph', 'p<n>', 'first' or 'random'
 ## Randomly choose 1 of the fastest 2, half, n, 1 or all live servers by latency.
 ## The response quality still depends on the server itself.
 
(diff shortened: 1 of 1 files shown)


lifenjoiner:
Ref:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/f2700874fd43d1531b025c93bbe6374c2d14846c/dnscrypt-proxy/config.go#L407-L430

jedisct1:
Thanks!

`p<n>` is something that doesn't really make sense. While it was implemented following a request in a GitHub issue, it was not documented because it may eventually be removed.

`DEBUG` is also intentionally not documented. This is really only useful for developers, and even then, only occasionally to debug specific issues. Debug mode can also change the server's behavior, for example to make execution deterministic, making the application completely insecure for actual usage.

lifenjoiner:
OK, got it.

But `p<n>` maybe help in a badly QoS network: user can set it listen several ports with 1 process, then with another local forwarder which can _query to all these ports and return the 1st response_. As the latency won't be updated before having a result, a greater than `count of server ports` of `n` is needed. `ph` may be too many to get the very slow servers. So if `p<n>` is removed, it can only be multi-process for this.

We definitely don't set `DEBUG` in daily use. But when debugging I always forget it. I also saw forgetting this in the issue dialogs while suggesting people to get logs ;p

jedisct1:
Well, we can keep `pn`.

But maybe not DEBUG. People freak out when they see suspicious things in logs, and report irrelevant issues. DEBUG is also a way to avoid this.

lifenjoiner:
`ph` may be too many ~~to~~ that will get the very slow servers.

`DEBUG`:
Yeah, people freak out of suspicious things.
I mean we should document this in somewhere instead of always memorizing it. When we need debug, we can set it right quickly. I don't know a better place than just a comment. I have been puzzled by why don't all `<2` work for at least 2 times, then went to read the source code 😂
BTW, on Windows, set a variable in normal console only persists in its session, it won't affect another non-child process. And setting a variable in console could be too complicated to the people freak out of suspicious things. How about other OS?

jedisct1:
> I mean we should document this in somewhere instead of always memorizing it.

On a Post-It™ 😂

lifenjoiner:
To be not confusing, which do you prefer, 'p\<n\>' or 'pn'?

What about:
```
## If you intend to set it to `< 2` to debug/dev, environment variable `DEBUG` is needed.
```
Do you think it's too easy for people to trigger it unexpectedly? There are 2 switches required. People should be responsible for what themselves do, rather than blame you. I personally hoping there can be a notice at hand to be friendly to super users and developers. Please~ One more try :joy: 

BTW:
Please take a look at https://github.com/DNSCrypt/dnscrypt-proxy/issues/1490
Seems you forgot to optimize it while adjusting the ratios.

jedisct1:
> To be not confusing, which do you prefer, 'p<n>' or 'pn'?

Maybe `p<n>` or even `p immediately followed by a number (ex: p5)`.

> Do you think it's too easy for people to trigger it unexpectedly

Not unexpectedly, but they are still going to complain if odd things are being printed.

lifenjoiner:
I have done all I can do

I will keep the comment about `DEBUG` in my own config file.

jedisct1:
Thank you!

Issue #1488, opened 2020-10-06
Adding an IP address to the whitelist has no effect


curiosityseeker:
## *Who* is the bug affecting?

Possibly everyone using an IP blacklist.

## *What* is affected by this bug?

False positives in IP blacklists cannot be bypassed.

## *When* does this occur?
Always.

## *How* do we replicate the issue?
1. Add, e.g., https://raw.githubusercontent.com/ktsaou/blocklist-ipsets/master/firehol_level1.netset as your IP blacklist (recently renamed to _blocked_ips_file_)
2. That list contains the IP address 81.169.145.105 as a false positive.
3. Add that IP address to the whitelist (recently renamed to _allowed_names_file_) and restart dnscrypt-proxy.
4. You'll observe that that IP address is still blocked.

## Expected behavior (i.e. solution)
As it seems that there is no specific IP whitelist available it should be expected that _allowed_names_file_ also allows for bypassing false positives in IP blacklists.

## Other Comments

I suggest to add an _allowed_ips_file_

PR #1492, opened 2020-10-13
add allowed ips plugin

ianbashford:
RE: issue #1488 
First cut at an Allowed IP plugin - I wasn't sure what you thought of the suggestion in that issue but it didn't seem unreasonable.  Of course if you have issues with the concept let me know... 

Notes:  I've begun to adopt 'allow' in the code. I've not refactored it all in order to keep the size of the PR reasonable, but this should be a step towards aligning the code with the external facing terminology.

WIP - tested with same IP in both blocked-ips.txt and allowed-ips.txt and shown it to provide the functionality.  I've not tested the log file or formats yet... thought I'd garner some feedback and see how it was received.


ianbashford:
Quick note about the implementation here
THe priorities are as follows:
AllowedName > BlockedName > AllowedIP > BlockedIp

The change here of course is just the insertion of AllowedIP
Otherwise I think the MR is in good shape now

jedisct1:
Thanks a lot Ian!

Is the `ltsv` log format actually unsupported? In the plugin code, it looks like it is handled.

ianbashford:
It should support `ltsv`  - I must be missing something as I can't see the block of code where I exclude it.  
`config.go`  looks OK doesn't it?  If you don't mind giving me a pointer, I'll correct whatever is misleading... 

<edit>  seems to output `ltsv` OK

jedisct1:
NVM, I was confused by the code in `config.go`.

ianbashford:
Oh -- I just had to look at that a few times....   I ended up running it to convince myself.

ianbashford:
Hi -- have you had a chance to take a look at this?
I you're OK with the concept I can rebase it against master -- if you're not so keen on the concept it's NP either... happy to discuss as well of course.
Any thoughts / comments welcomed.

jedisct1:
If it can be useful, yes, totally!

If you could rebase it, that would be wonderful. I'll merge it right after.

ianbashford:
I'm not having fun with the rebase...   I'll resubmit a new PR

PR #1518, opened 2020-11-21
Add fake IP plugin

rikkihamano:
For some sites with CDN, sometimes we want to use the specific IP to achieve a faster experience.


jedisct1:
That looks to address a very niche problem, the most general cases being already addressed with cloaking and clientsubnet. This is also going to be very fragile, as IP addresses returned by CDNs change all the time.

Finally, redirecting connections to an IP address to another IP address is more of something to be done at firewall level rather than DNS.

So, this is great, and maybe useful, but I'm not sure that it wouldn't bloat the code with something that very few people will eventually use.

Maybe this is a motivation to have the ability to be able to support external plugins, even if Go is not really the best language for this.

PR #1523, opened 2020-11-24
Change example forwarding rule to match recommended .home.arpa TLD

petercooperjr:
The ".home" TLD was proposed at one point, and while it's probably not going to actually ever get delegated it's not best practice to just start using your own TLD. The .home.arpa domain has been specifically set aside for use in home networks (RFC 8375) and is probably the better example to put here.


jedisct1:
`.home` [looks safe](https://www.theregister.com/2018/02/12/icann_corp_home_mail_gtlds/)!

This list is mainly designed to cope with what routers already do :(

And if people have to choose their network name, there is very little chance that anyone would choose `.home.arpa` while they could choose `.lan` instead.

Let's add `.home.arpa` to the list, but `.home` deserves to stay, as it is probably far more common.

PR #1527, opened 2020-11-27
download mirror download.dnscrypt.net

mibere:
Files are locally hosted on download.dnscrypt.net. A cronjob updates the files every 3 hours, source is https://download.dnscrypt.info
download.dnscrypt.net has IPv4 and IPv6, DNSSEC, HTTPS


jedisct1:
Thank you!

mibere:
FYI: Mirror is still valid...but currently down. My fault 🤦‍♂️ Hit a Let's Encrypt rate limit while testing new TLS configurations; didn't know that there is a limit. Ouch!

PR #1534, opened 2020-12-05
Add options to hide domain names and client IPs in the query log

plumbeo:
This PR adds two new options to the query log plugin, `hide_client_ip` and `hide_domain_name`, that replace respectively the client IP address and the queried domain name with `***` in  the query log file.

My use case is that I like having the query log available to do some statistical analysis but I'd prefer not to have a permanent record of who looked for what.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -168,9 +168,11 @@ type SourceConfig struct {
 }
 
 type QueryLogConfig struct {
-	File          string
-	Format        string
-	IgnoredQtypes []string `toml:"ignored_qtypes"`
+	File           string
+	Format         string
+	IgnoredQtypes  []string `toml:"ignored_qtypes"`
+	HideClientIP   bool     `toml:"hide_client_ip"`
+	HideDomainName bool     `toml:"hide_domain_name"`
 }
 
 type NxLogConfig struct {
@@ -490,6 +492,8 @@ func ConfigLoad(proxy *Proxy, flags *ConfigFlags) error {
 	proxy.queryLogFile = config.QueryLog.File
 	proxy.queryLogFormat = config.QueryLog.Format
 	proxy.queryLogIgnoredQtypes = config.QueryLog.IgnoredQtypes
+	proxy.queryLogHideClientIP = config.QueryLog.HideClientIP
+	proxy.queryLogHideDomainName = config.QueryLog.HideDomainName
 
 	if len(config.NxLog.Format) == 0 {
 		config.NxLog.Format = "tsv"
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -432,6 +432,16 @@ cache_neg_max_ttl = 600
   # ignored_qtypes = ['DNSKEY', 'NS']
 
 
+  ## Do not log the clients' IP addresses.
+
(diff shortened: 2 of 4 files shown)


jedisct1:
Hi!

This is a bit of a niche use case, so I'd be a bit reluctant to adding a feature that very few people would actually use.

Plus, if such a feature was introduced, it should be generalized, not just implemented for two specific properties of one specific log file.

The good news is that all of this and more can already be done by [logging to a pipe](https://github.com/DNSCrypt/dnscrypt-proxy/wiki/Logging#custom-log-format--log-processors).

plumbeo:
I pushed a cosmetic change and test #6 is failing, but it’s failing on the master branch too so I suppose there’s something else going on.

Yes, there are other ways to transform the logs, but they all require a lot more infrastructure to work reliably: something needs to ensure that the named pipe exists before dnscrypt-proxy runs, then something else needs to be written to process the output and save it somewhere else, and then something else again needs to ensure that all this is running whenever dnscrypt-proxy is running, and finally something else needs to rotate the new log file because the built-in log rotation mechanism doesn’t work anymore. 

It’s a lot of work for a basic privacy feature and it’s not that simple to do, to do correctly and to integrate with the distro-provided configurations or to implement in more limited environments, while this PR takes care of everything with two options and a dozen new lines of code.

If you think it would be better to generalise is I can look at it. Where else do you think it could be useful? I was thinking of the NX log but I may have missed something else.

jedisct1:
> Yes, there are other ways to transform the logs, but they all require a lot more infrastructure to work reliably.

Come on, you're smart, you can run a shell script.

Doing these transformations externally is a way better option IMHO. It offers far more flexibility.

plumbeo:
I know I can, I’ve done a lot worse, but I’m not really looking forward to copying scripts and custom systemd unit and overrides in every machine where I deploy dnscrypt-proxy :) 

At that point it becomes easier to compile a modified binary, copy it around and pin the installed package so the distro doesn’t overwrite it, or just give up and create a custom deb and place it in a repository, which I would still need to maintain and keep up to date and add everywhere, so that would probably end up being even more work :/

Would you be open to a more general PR, with global options to hide domain names and client IP addresses in all the logs where they’re printed out, with the possibility to override the values on each single plugin? Assuming that it’s possible, I didn’t check the code yet.

jedisct1:
A generic implementation would rather be the ability to define a custom error log format for every log.

But really, since no one ever asked for such a feature before, and since that would still be insufficient for people using log processors, I don't think this is a good idea.

And sharing external processing scripts would be more useful, as anyone could easily customize or extend them to their needs.

plumbeo:
> A generic implementation would rather be the ability to define a custom error log format for every log.

I can’t promise anything but if I came up with a new PR implementing a custom log format (something like the traditional Apache-style LogFormat, with `%` directives and all), assuming that the quality of the code is good enough, would it be considered for inclusion?

> But really, since no one ever asked for such a feature before, and since that would still be insufficient for people using log processors, I don't think this is a good idea.

I’m a little surprised by that, IP addresses on their own are PII (personally identifiable information) and the combination of addresses and domain names must be treated even more carefully by every business subject to GDPR, so something like what I implemented in this PR was a limited but quick fix. I suppose everybody’s either not logging anything or filtering the logs somewhere else (or simply didn’t notice that the query log is on by default in several major distributions, certainly unlikely but it wouldn’t be completely surprising).

PR #1564, opened 2020-12-28
Run from in memory cache updates

### LINKED ISSUE #1529: Avoid using outdated cache files
## Is your feature request related to a problem?

As we could [read](https://www.reddit.com/r/dnscrypt/comments/jvq6je/dnscryptproxy_users_please_check_if_your_cache/gcmsi47?utm_source=share&utm_medium=web2x&context=3) at Reddit there is the problem that there are users with very outdated cache files.

My server has a [new IP](https://github.com/DNSCrypt/dnscrypt-resolvers/commit/f15f0da39b9afa608eeb28c86f55c9d05fc7a001) for 27 days now, but there are still several thousand queries on the old IP per day (the old server is still running).

As server operator I want that the client (dnscrypt-proxy) respects my changes at some point.

## Describe the solution you'd like

The client (dnscrypt-proxy) should in some way force an update of the files (or in other words: the client should force the user), or completely refrain from answering requests if the files are too old, e.g. older than 20 days.

## Who will that feature be useful to?

It is beneficial for users and server operators.

/cc @ianbashford

ianbashford:
@mibere @jedisct1 @welwood08    as discussed in #1529  

Fix 1: maximum refresh delay of 1 week + example file comments updated
(see config.go  - loadSource )

Fix 2: 'prefetchSources' timer loop now also updates registered servers
I needed to refactor part of 'loadSource' into proxy.go so it can be called from the timer loop; that's implemented in 'updateRegisteredServers() '.   Even if the cache area can't be written, the in memory downloaded files are used.

Implementation:
The proxy.sources.bin (ary) representation of each file is already updated from the regular timer. 
updateRegisteredServers is triggered on the prefetch timer and this updates the serverInfo 'registeredServers'
NOTE: it always updates all of the registered servers in serverInfo - possible improvement to only call registerServer on a change
NOTE: it doesn't remove servers - it updates the stamps for current servers or adds new servers
NOTE: implementation should work for new relays too, but as they'd need to be configured in the file, this is likely superflous

I found the easiest way to test was to comment out the signature check, and have a couple of files on a local webserver I flipped between.

Comments welcomed...


ianbashford:
I've added a commit to cover the source prefixes now.

ianbashford:
Two more small commits.  The first just stops the repeat of register server which has already been called in 'updateRegisteredServers()'
The second isn't directly related to this PR -- refreshServer had its own 'append' to serversInfo.registeredServers()  - this bypassed the logic in the registerServer function that checks for duplicate names.   I've just reused that function to avoid the duplicate

dnscrypt-proxy/config.go
```diff
@@ -750,22 +768,13 @@ func (config *Config) printRegisteredServers(proxy *Proxy, jsonOutput bool) erro
...
-	if config.SourceRequireNoLog {
-		requiredProps |= stamps.ServerInformalPropertyNoLog
-	}
-	if config.SourceRequireNoFilter {
-		requiredProps |= stamps.ServerInformalPropertyNoFilter
-	}
 	for cfgSourceName, cfgSource_ := range config.SourcesConfig {
 		cfgSource := cfgSource_
-		if err := config.loadSource(proxy, requiredProps, cfgSourceName, &cfgSource); err != nil {
+		if err := config.loadSource(proxy, proxy.requiredProps, cfgSourceName, &cfgSource); err != nil {
```
welwood08: Looks like `requiredProps` doesn't need to be passed into this function at all now since the usage has moved to the new `updateRegisteredServers` function.

ianbashford: Ack.  Commit pushed to PR

jedisct1:
Awesome work, Ian! Thank you so much! (and thank you Will for the review!)

mibere:
I'm not a Git expert, but does the commit also includes https://github.com/DNSCrypt/dnscrypt-proxy/pull/1564/commits/d6011d7f53ab0ab49df21735362e0a068e5b5588 ?

I ask as https://github.com/DNSCrypt/dnscrypt-proxy/blob/master/dnscrypt-proxy/example-dnscrypt-proxy.toml doesn't inlcude the changes.

jedisct1:
@mibere 

I'm confused here. Doesn't https://github.com/DNSCrypt/dnscrypt-proxy/blob/master/dnscrypt-proxy/example-dnscrypt-proxy.toml include the changes for you? I don't see any `blacklist` instance here, everything seems to have been replaced with `blocklist`.

mibere:
Don't know what was wrong ;-) All seems to be correct.

jedisct1:
Looks like static entries are now [ignored](https://github.com/Haraade/dnscrypt-proxy/commit/0557253389e0c7de71eddcfff25ff0667bcc6a60#commitcomment-45576977).

/cc @ianbashford

ianbashford:
> Looks like static entries are now [ignored](https://github.com/Haraade/dnscrypt-proxy/commit/0557253389e0c7de71eddcfff25ff0667bcc6a60#commitcomment-45576977).
> 
> /cc @ianbashford

Checking...

ianbashford:
I see it.  
proxy.updateRegisteredServers() is called _before_  statics are loaded.
I've just tested moving the call after and it picks up static only configs.
I'll send the update shortly

ianbashford:
I've pushed it into this PR to make it easier to find.
I need to clean my tree a little (which is I think why the PR was a little messy), so I'll create new PRs if anything else comes up.

ianbashford:
Fresh PR incoming in https://github.com/DNSCrypt/dnscrypt-proxy/pull/1568

jedisct1:
It this thread-safe?

`updateRegisteredServers()` can add entries to `proxy.registeredRelays` and `proxy.registeredServers`, so I'm a little bit worried about race conditions.

ianbashford:
Actually I didn't specifically tackle thread safety -- after the config call at startup, it's only called by the timer thread
```
grep updateRegistered *.go
config.go:	proxy.updateRegisteredServers()
proxy.go:			proxy.updateRegisteredServers()
proxy.go:func (proxy *Proxy) updateRegisteredServers() error {
```
Is there a specific use case you have in mind where it might be called by > 1 thread?
I can take a look at adding thread safety to it.

ianbashford:
OK -- I've taken a look.  I think this is safest with a `lock.Lock()` in `proxy.updateRegisteredServers()`    to avoid `serversInfo`  trying to read while `updateRegisteredServers` is trying to write.
I think I understand that as the troublesome usecase...  hopefully it's a fix to any other issues as all the writing is done in that function

ianbashford:
Actually, are we better to create `proxy.serversInfo.registeredRelays` in the same pattern as `proxy.serversInfo.registeredServers`  and keep the single lock in serversInfo?
There's a bit of refactoring then to duplicate `registerServer()` into `registerRelay()`, then apply the lock each time before reading  []RegisteredRelays...   
I see a lot of additional reads for the relay selection and it might be best contained in serversInfo...

PR #1568, opened 2021-01-02
register servers after loading statics

ianbashford:
Statics were not registered correctly at launch time.  
Moves the first call to register servers until after the statics are loaded


jedisct1:
Awesome, thank you so much Ian!

PR #1569, opened 2021-01-02
add tests for statics

ianbashford:
I've added a couple of CI tests for static only configuration.
I've used the stamp for your scaleway-fr server - hope that's OK.  I thought it best to choose a stable server under your control...


jedisct1:
Good idea, thank you!

PR #1579, opened 2021-01-04
Threadsafe update

ianbashford:
@jedisct1 
First commit moves registeredRelays into serversInfo (analagous to registeredServers)
Second commit adds locks around each copy taken of registeredRelays in serversInfo.

Do you think this might cover it?
Happy for pointers or alternative solutions... 
This will need a careful code review - Is it bad form to start each function with a `defer serversInfo.Unlock()` to ensure as each function exits we never leave a lock in place?

I'm currently testing this with 3 server_names and a '*' in the routes

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -352,6 +352,9 @@ func (proxy *Proxy) updateRegisteredServers() error {
 	for _, registeredServer := range proxy.registeredServers {
 		proxy.serversInfo.registerServer(registeredServer.name, registeredServer.stamp)
 	}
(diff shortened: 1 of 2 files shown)


jedisct1:
Thanks a lot Ian!

It looks good.

I'm going to run a build with `-race` for quite some time to see if anything happens, but I think we're pretty safe now!

ianbashford:
Great stuff - I've pulled a new cut as well which I'll keep on long term test.  I've got it with a * in routes and I'm accepting all servers, so we'll see if we get any issues.
If you can tell me how to '-race'  I'll give it a go too

jedisct1:
Just compile with `go build -race` :)

PR #1613, opened 2021-02-09
Reduce relay sources downloading

lifenjoiner:
Save some resources for the servers :)

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -787,6 +787,9 @@ func (config *Config) printRegisteredServers(proxy *Proxy, jsonOutput bool) erro
 
 func (config *Config) loadSources(proxy *Proxy) error {
 	for cfgSourceName, cfgSource_ := range config.SourcesConfig {
+		if strings.HasPrefix(cfgSourceName, "relays") && (proxy.routes == nil || len(*proxy.routes) == 0) {
+			continue
+		}
 		cfgSource := cfgSource_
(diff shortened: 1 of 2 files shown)


jedisct1:
Mmm... I have mixed feelings about that change.

Relays are not necessarily meant to be in a dedicated file. It would be fine to combine everything in a single file.

That would also be a breaking change, and the prefix requirement can introduce unnecessary confusion (also `prefix` by itself is confusing since it has its own meaning in these sections).

Finally, it's nice to keep all the files up to date, even if some of them are not currently used by a given configuration.

lifenjoiner:
> Relays are not necessarily meant to be in a dedicated file.

That's why I used `HasPrefix`, considering if there is any other user specified relays.

> It would be fine to combine everything in a single file.

This is hard to cover. But, yes, that's possible.

Just a suggestion that better than an issue ;)
It's OK to close this PR if you don't think it's practicable.

jedisct1:
This is practicable, but it slightly complicates the code and the configuration (and we'll have to fit the OdOH relays in the picture at some point) without any real user benefit. Configuration downloads don't require a lot of traffic (these files are very small and the download is gzipped), so that would mitigate a non-existent issue.

lifenjoiner:
The PR works only when made `relays*` the reserved source names for relays.
Currently, it is free in the `sources` definition. I have no better idea.

I guess most people use the example configuration, but have no idea how much it costs. If it doesn't matter, that will be good.
When it really matters, comment off the relays section in the `.toml` file is another option, costs less.

PR #1632, opened 2021-03-01
Fix domain blocklist config example

kazzarin:
`blocklist` doesn't seem to be a valid key
https://github.com/DNSCrypt/dnscrypt-proxy/blob/1a82786e078d217caf40e06553a8f95398ffde4a/dnscrypt-proxy/config.go#L64-L65

so I changed it to the values from the main example config
https://github.com/DNSCrypt/dnscrypt-proxy/blob/09866acdb522b51ef1bc658a19c2ab84efafca67/dnscrypt-proxy/example-dnscrypt-proxy.toml#L506-L510


jedisct1:
Thank you!

PR #1653, opened 2021-03-29
Add ODoH support.

chris-wood:
Integration tests make use of external servers right now. We can certainly change that if we want to make them more robust. I'm definitely open to any and all feedback for how we might improve this or increase test coverage!


dnscrypt-proxy/proxy.go
```diff
@@ -678,6 +678,42 @@ func (proxy *Proxy) processIncomingQuery(clientProto string, serverProto string,
...
+				responseBody, responseCode, _, _, err := proxy.xTransport.ObliviousDoHQuery(targetURL, odohQuery.odohMessage, proxy.timeout)
+				if err == nil && len(responseBody) > 0 && responseCode == 200 {
+					response, err = odohQuery.decryptResponse(responseBody)
+					if err != nil {
+						dlog.Error("Failed to decrypt response")
+						response = nil
+					}
+				} else if responseCode == 401 {
+					dlog.Notice("Forcing key update")
+					go proxy.serversInfo.refresh(proxy)
```
alisonatwork, line 702: @chris-wood Just looking through this change and reading up on ODoH in general, it's interesting stuff!

Regarding this line of code - I think we don't yet have anywhere in the code that refreshes the whole server list after we started up. Will running this in a goroutine potentially kick off a long or heavyweight background process for people with a lot of servers configured, and won't it also trigger refreshes for non-ODoH servers? I'm not sure what the goal is here, but if it is just to refresh a single server, perhaps there could be a way to just trigger a (synchronous?) refresh on that one. I might also just be misunderstanding how this works, it's been a while since I had a look at the dnscrypt-proxy source.

chris-wood: Good point, @alisonatwork! I went with this solution since it reused as much of the existing update machinery as possible, but if that's a possibly long and heavyweight background process, perhaps we should only update just this specific server info. @jedisct1, what do you think?

jedisct1: Hi!

Yes, updating the whole list indeeds doesn't seem optimal. But what worries me more about this is that we may have some thread safety issues here. `refresh()` is not expected to be called from multiple threads simultaneously.

So, if you can find a way to only refresh the server's key instead that would be much better.

Entire refreshes happen regularly in the background. I'm assuming that after a key rotation, the previous key still remains usable for some time? In that case, a 401 response should virtually never happen, so it may be fine to do a synchronous update.

chris-wood: > Entire refreshes happen regularly in the background. I'm assuming that after a key rotation, the previous key still remains usable for some time? In that case, a 401 response should virtually never happen, so it may be fine to do a synchronous update.


Yeah, a synchronous, single server update here seems best. :-) Do you want to take this, or shall I?

chris-wood: FWIW, I proposed #1671 to try and address this issue. Thanks again for flagging it, @alisonatwork. :-)

alisonatwork: Thanks for addressing this @chris-wood! I know it's just a small corner case, but the nicer we can make support for ODoH out of the box, I think the more likely people will adopt it.

PR #1663, opened 2021-04-05
replace TrimFunc(s, IsSpace) with TrimSpace for ASCII optimization

alisonatwork:
Just a small cleanup that a linter turned up. `TrimSpace` is possibly a bit faster than `TrimFunc` with `IsSpace`, because if the string being trimmed only has ASCII spaces, it doesn't need to do the unicode check.


jedisct1:
Nice!

Thank you Alison!

PR #1664, opened 2021-04-07
Temporarily disable `Geoffrey Frogeye's trackers list`

IceCodeNew:
For `hostfiles.frogeye.fr` has been offline for hours.

Refer:
a. https://port.ping.pe/hostfiles.frogeye.fr:443
b. https://i.ping.pe/h/M/img_hMz19G0Y.png

### DIFF
--- a/utils/generate-domains-blocklist/domains-blocklist.conf
+++ b/utils/generate-domains-blocklist/domains-blocklist.conf
@@ -75,7 +75,7 @@ https://raw.githubusercontent.com/nextdns/cname-cloaking-blocklist/master/domain
 # https://adguardteam.github.io/AdGuardSDNSFilter/Filters/filter.txt
 
 # Geoffrey Frogeye's block list of first-party trackers - https://hostfiles.frogeye.fr/
-https://hostfiles.frogeye.fr/firstparty-trackers.txt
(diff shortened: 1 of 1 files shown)


jedisct1:
Looks like it's back :)

IceCodeNew:
I suggest we add a To-do for `generate-domains-blocklist.py` is not dealing with network error well.
While generating a blocklist, any source which has failed to fetch could lead to an empty list. Such a result is not acceptable for me.

jedisct1:
If you need to use the output immediately, the common way to do it is to use a temporary file, that gets renamed on success:

```sh
generate-domains-blocklist.py -o list.tmp && mv -f list.tmp list.txt
```

This also ensures that the file is always completely written.

IceCodeNew:
> If you need to use the output immediately, the common way to do it is to use a temporary file, that gets renamed on success:
> 
> ```shell
> generate-domains-blocklist.py -o list.tmp && mv -f list.tmp list.txt
> ```
> 
> This also ensures that the file is always completely written.

You are right. I definitely need to clean up some old codes.
This problem could be fixed on my side.

IceCodeNew:
@jedisct1 `hostfiles.frogeye.fr` is offline again :-(

PR #1671, opened 2021-04-17
Synchronously update the target configuration upon failure.

chris-wood:
If the target _fails_ to update, nothing else is done. This relies on the load balancer to try and resolve the query via some other server. This is an exceptional case, so we could definitely do something different here, e.g., remove the target from the list of registered servers. 

@jedisct1, do you think something different should be done here? I'm totally open to alternate recovery flows!


jedisct1:
That should be good!

Just one small nit: could you make the warning messages a little bit more descriptive?

Maybe add the server name to `Forcing key update` and `Key update failed`.

jedisct1:
@alisonatwork Does that look good to you?

chris-wood:
> Maybe add the server name to Forcing key update and Key update failed.

Good idea! Done.

PR #1694, opened 2021-04-29
Upgrade request to https

arch-btw:
Upgrade request to https

### DIFF
--- a/utils/generate-domains-blocklist/domains-blocklist.conf
+++ b/utils/generate-domains-blocklist/domains-blocklist.conf
@@ -54,7 +54,7 @@ https://pgl.yoyo.org/adservers/serverlist.php?hostformat=nohtml
 # https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt
 
 # Sysctl list (ads)
-# http://sysctl.org/cameleon/hosts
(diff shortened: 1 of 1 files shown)


jedisct1:
Unfortunately this is intentional. Their TLS configuration is a little bit atypical.

TLS 1.2 connections are not supported, and while we may assume that Python is going to use TLS 1.3 on every platform, this is likely not the case. Even connecting to that website from macOS with the `curl` command is currently impossible.

arch-btw:
Hey, thank you for your reply and for your hard work.

That's really unfortunate that they have an atypical TLS configuration.

In that case it's better not to merge. 

Thank you again!

arch-btw:
I just noticed that the hosts file is depreciated and no longer updated:



On their website: http://www.sysctl.org/cameleon/

Maybe a solution is to use the archived version on archive.org?

https://web.archive.org/web/20201112015556if_/http://sysctl.org/cameleon/hosts

If you rather not I totally understand :-)

jedisct1:
Whoops, thanks for the heads up!

If it is not maintained any more, it's probably better to remove it. Legacy lists tend not to be very useful and have a lot of false positives.

PR #1732, opened 2021-06-05
Add missing return at end of function

syphyr:
./odoh.go:90:1: missing return at end of function

fixes: "Limit the number of ODoH target configs"


jedisct1:
Thanks!

A slightly different fix was made since we probably don't want to return an error here and just keep the first configurations instead.

PR #1764, opened 2021-06-26
Updated README.md

amr3k:
Made the logo centered


jedisct1:
Hi,

That file currently doesn't have any HTML markup. It's only text, that can be easily read offline as text, without being rendered on GitHub and viewed with a web browser.

Unless necessary for semantic reasons, I'd rather keep it that way that start mixing HTML sigils into the documentation.

That would be different on a web page, but in a Markdownfile, it's okay for the logo not to be centered :)

PR #1765, opened 2021-06-27
Update example-blocked-ips.txt

jacob755:
add dns rebinding protection to the example file so everyone know about it


jedisct1:
Hi, and thanks!

I'm not really convinced that this is a good idea or even necessary, though.

First, it will break valid setups. Devices from your local network won't resolve any more. SPF-based spam filters won't work any more either. 

Reliable DNS rebinding attacks rely on short TTLs. But `dnscrypt-proxy` can enforce a minimum TTL, and does it in the default configuration. 

So, blocking these IPs would cause more trouble than solve actual issues. Even as an example, it would be probably better to keep these entries commented.

Maybe this would be better done as a new plugin, that prevents cached records to be replaced with internal IP addresses if they previously were external IP addresses.

It's probably fine to keep that documented in the wiki, as we can document the caveats and keep blocking internal IPs a manual and intentional operation for people explicitly looking for this. But I would honestly never use or recommend such a setup.

jacob755:
thanks for info. i never know about this ability so thought  should be in file. for single pc access internet work for me.
if you can make plugin better  if not how about just add comment line about it?
`#Extra protection with dnscrypt-proxy see https://github.com/DNSCrypt/dnscrypt-proxy/wiki/Filters#dns-rebinding-protection`
and updating document abut breaking area
otherwise close it.
thanks

PR #1770, opened 2021-07-02
Let Dependabot update GitHub Actions

jauderho:
This will make it so that Dependabot is able to submit pull requests to keep GitHub Actions up to date.


jedisct1:
Thank you!

PR #1822, opened 2021-08-14
Minor cleanup

a1346054:
Most importantly, make sure github (and everyone else) recognizes the license properly as ISC license.


jedisct1:
Thank you!

The space after `#!` is intentional, though. I would really like to keep it that way.

Old unix junkies may know why :)

a1346054:
Understood, it is a matter of personal preference.

I did however do a scan of all script files in my $PATH and checked if the first line begins with #!/ or #! / and the vast majority of them are #!/ by now. So wanted to unify the rest.

You can pick the other commits from the pull request, or I can revert that one commit, or rebase it completely. Up to you.

For me, the most important part is having the proper license file, because that's what was preventing the use of dnscrypt-proxy at $dayjob... really...

a1346054:
Rebased without the #! changes.

PR #1835, opened 2021-08-19
Add captive portal entry for dual stack setups

ghost:
For users with a dual stack set up at home (IPv4+native IPv6), Windows 21H1 will report that there is no Internet connection if this entry is missing from their captive portals file.

This pull request fixes the issue by adding the required entry to the sample file.


jedisct1:
Thank you! 😊

Issue #1654, opened 2021-03-30
TTL limits not honored


ghost:
Starting from 2.0.45 TTL limits in the TOML aren't properly honored anymore:

If one resolves a domain and then sends a 2nd query during the cache period the actual TLL can be much longer then the expected one, actually up to twice as long.
E.g. resolve 'github.com' with default TOML says it has a 40min TTL. Now send a 2nd query after 39min, so it's served from the cache, again having a TTL of 40min. You end up with an actual TTL of 39min + 40 min = 79min, instead of the intended 40 min.

If that's the new intended behaviour, you should rename TTL (e.g. TTBC = time to be cached) to avoid confusion and lower the current TTL values in the TOML. We then would also need a new way how to limit the actual TTL.

Verified on Windows and Linux. Not tested, but Mac treats TTL differently, right?

This is a direct consequence of bug [#1624](https://github.com/DNSCrypt/dnscrypt-proxy/issues/1624).

Ah, and btw. there's still one mention of the term 'whitelisting' in 'dnscrypt-proxy.log' when using an allowed-names list.

jedisct1:
Can you please open a discussion about this first?

Thank you.

PR #1895, opened 2021-09-23
Update TTL for cached queries before returning them

### LINKED ISSUE #1624: Cached responses had a non-decrementing TTL before version 2.0.45
## *Who* is the bug affecting?

Users of dnscrypt-proxy.

DNS cache and TTL does not behave as it did in prior versions. TTL does not decrement on the second query for the same domain name and queries after that show erratic TTL behavior.

## *When* does this occur?

It is most easily seen between the first and second queries for the same domain name with dnscrypt-proxy cache enabled.

## *Where* does it happen?

Version 2.0.45: macOS, linux-arm, linux-arm64

1. install dnscrypt-proxy version 2.0.45 from GitHub releases. (reproduced on macOS, linux-arm, and linux-arm64)
2. query a domain that will have it's TTL set according to the min-TTL defined in the .toml file.
3. see in the response that it's TTL is set to that min-TTL value (e.g. 2399).
4. make a 2nd query for the same domain name after a few seconds (or minutes) have passed.
5. see in the response for the second query that it's TTL still shows as 2399 rather than the expected decremented TTL, but it has responded from cache according to the query log file.
6. immediately make a 3rd query and see that it's TTL "jumps" down to a decremented TTL in accordance with how much time has passed since the 1st query.

1. The 1st query for a domain name would be cached for it's TTL or the min-TTL defined in the .toml file. (e.g. 2399 seconds).
2. After waiting 90 seconds, a 2nd query for the same domain name would then show a decremented TTL of original TTL - 90. (e.g. 2309 seconds in this example)

## Other Comments
Since updating to version 2.0.45 I have observed a change in cache behavior in comparison to prior versions. I am not sure if this is known or intended, but I thought I would ask. When a DNS query is made I would expect it to be cached for the TTL and then decrement in real time. On version 2.0.45 the first query shows a TTL of 2399 in accordance with the default .toml file. Then the second query also shows a TTL of 2399 no matter the amount of time that has passed in between the first and second query. If I waited 100 seconds between queries, for example, I would expect the TTL of the second query to have a TTL of 2299. You can tell that it is responding from cache, however, because the response time is much lower than the initial query and it is indicated in the query log. The third query will then "jump" to the expected TTL in relation to when the first query was made. As more queries for the same domain name are made, the response TTL continues to be odd. As seen in one of my screenshots, I made two queries about 103 seconds apart, however the TTL only decreases by 3.

I have re-run my tests using version 2.0.44 and it does not exhibit this behavior. The TTL decreases from the first to second query and also reliably decreases in accordance with the amount of time that has passed between queries.

I don't believe this is a "just me" issue as I have reproduced this behavior on the release binaries for macOS, linux-arm (raspberry pi), and linux-arm64 (router). I have also compared past versions and this is new as of 2.0.45. Version 2.0.44 behaves as expected.

**Screenshots:**

Screen Shot 2021-02-21 at 9 45 21 PM

Screen Shot 2021-02-21 at 9 40 14 PM

Screen Shot 2021-02-22 at 8 43 48 AM

ghost:
Supposed to fix #1624, #1654, #1894, maybe #1889 and possibly others.

Many thanks to KilometerM who did a great job describing the issue.

### DIFF
--- a/dnscrypt-proxy/plugin_cache.go
+++ b/dnscrypt-proxy/plugin_cache.go
@@ -78,6 +78,8 @@ func (plugin *PluginCache) Eval(pluginsState *PluginsState, msg *dns.Msg) error 
 		return nil
 	}
 	cached := cachedAny.(CachedResponse)
+	
+	updateTTL(&cached.msg, cached.expiration)
 
 	synth := cached.msg.Copy()
 	synth.Id = msg.Id
@@ -89,8 +91,6 @@ func (plugin *PluginCache) Eval(pluginsState *PluginsState, msg *dns.Msg) error 
 		pluginsState.sessionData["stale"] = synth
 		return nil
 	}
-
-	updateTTL(&cached.msg, cached.expiration)
 
 	pluginsState.synthResponse = synth
 	pluginsState.action = PluginsActionSynth


jedisct1:
Good catch, thanks!

A cached value is expected to be read-only, so we should update `synth` instead of `cached`. Done in 75e917a.

Thanks again!

ghost:
Thanks for fixing. How do we want to handle stale TTL or is it handled elsewhere?

PR #1901, opened 2021-09-25
Revert "Set standard stale ttl"

jedisct1:
Reverts DNSCrypt/dnscrypt-proxy#1900


ghost:
Why?

edit: Ah, I see, even better.

PR #1993, opened 2022-01-19
systemd: use constants and update status on ready

cobratbq:
Systemd-notify signaling indicates the status of dnscrypt-proxy when
starting as 'Type=notify' systemd service. However, the status is not
updated when initialization completes, instead it always shows
"Starting". Now fixed.


jedisct1:
This is great!

Thank you so much for this!

PR #2003, opened 2022-01-23
Download mirror dnscrypt.net removed

mibere:
The server is no longer available.


jedisct1:
Thank you for having provided that mirror, @mibere !

PR #2014, opened 2022-02-01
Update domains-blocklist.conf

ghost:
Re-added GameIndustry list which was removed because of an outdated link.

### DIFF
--- a/utils/generate-domains-blocklist/domains-blocklist.conf
+++ b/utils/generate-domains-blocklist/domains-blocklist.conf
@@ -149,3 +149,6 @@ https://hosts.oisd.nl/basic/
 
 # Block spying and tracking on Windows
 # https://raw.githubusercontent.com/crazy-max/WindowsSpyBlocker/master/data/dnscrypt/spy.txt
+
+# GameIndustry.eu - Block spyware, advertising, analytics, tracking in games and associated clients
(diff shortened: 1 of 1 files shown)


remyabel2:
This has way too many false positives. I was getting OCSP errors and saw that it was blocking not only essential mozilla services but innocuous stuff like bitbucket and duckduckgo.

jedisct1:
Thanks for your feedback, @remyabel2 !

Well... until it becomes more reliable, leaving it off the example list is probably the best thing to do, then.

ghost:
Well, this is a special purpose list mainly for stopping games and software from phoning home while accepting any breakage this may cause.
So the mentioned issues aren't false positives and it won't get more reliable. Therefore it might actually make sense to keep it off the example list.

remyabel2:
Try this instead: https://github.com/KodoPengin/GameIndustry-hosts-Template/blob/master/Gaming-Full-Template/hosts

I didn't test it because it has too many entries but it doesn't have the problematic entries from the other file.

> Well, this is a special purpose list mainly for stopping games and software from phoning home while accepting any breakage this may cause.

It definitely is a false positive. The main template blocks Firefox's ability to verify HTTPS certificates with OCSP so many sites will just not load at all. If they're trying to block telemetry, it would be better to use a user.js file instead.

PR #2015, opened 2022-02-02
Remove README.md "Contribute" link as the target doesn't exist

PeterDaveHello:
CONTRIBUTING.md doesn't exist at all 😅


jedisct1:
Good catch, thank you!

PR #2029, opened 2022-02-08
Use the same format logic throughout the document

d3cim:


jedisct1:
Thank you!

But maybe instead of removing indentation, it would be better to indent everything?

It increases readability, especially with long sections.

d3cim:
Ok, I got the logic now, after a look I think these are the correct changes, since:

`[query_log]`, `[nx_log]`, `[blocked_names]`, `[blocked_ips]`, `[allowed_names]` and `[allowed_ips]` are not subcategories of anything, unlike:

`[schedules.'time-to-sleep']` and `[schedules.'work']` for `[schedules]` or `[static.'myserver']` for `[static]`, must not be doubly spaced from the beginning of the margin I think.

`[schedules.'time-to-sleep']`, `[schedules.'work']` and `[static.'myserver']` are now intended instead.

ghost:
The title of this commit is now less true than before. This needs fixing.

d3cim:
Look at the [raw file](https://raw.githubusercontent.com/DNSCrypt/dnscrypt-proxy/761dd190f24e168d6b4a9dfb71bd7c4c2456c72e/dnscrypt-proxy/example-dnscrypt-proxy.toml), it seems fine to me.

PR #2034, opened 2022-02-09
Minor cosmetic toml changes

ghost:


dnscrypt-proxy/example-dnscrypt-proxy.toml
```diff
@@ -704,7 +704,7 @@ format = 'tsv'
 
   ### Quad9
 
-  # [sources.quad9-resolvers]
+  # [sources.'quad9-resolvers']
```
jedisct1: That one looks really weird. Why the quotes?

ghost: I think so, too. I did it to be compliant with all other subsections having quotes as well. Personally I would rather remove them all, though.

edit: updated toml

jedisct1:
Thank you!

PR #2037, opened 2022-02-10
Algebraic simplification

ghost:


jedisct1:
The factors are probably suboptimal. They were meant to be tweaked until we can find something that better fits most use cases.

That simplification doesn't convey the original intent any more, and it probably doesn't change much to the compiled code.

So I'd rather keep the expression as-is.

PR #2040, opened 2022-02-11
Improve estimator

ghost:
Just a concept:

1. Always partially sort because it should always be necessary (maybe except first time)
2. Sort from back to front to assure lowest RTT is always at the first place
3. Use `rtt.Set` instead `rtt.Add` to set candidate RTT in an easy way (e.g. the midpoint)

When do negative RTTs occur and how should they be handled?

### DIFF
--- a/dnscrypt-proxy/common.go
+++ b/dnscrypt-proxy/common.go
@@ -101,20 +101,6 @@ func Max(a, b int) int {
 	return b
 }
 
-func MinF(a, b float64) float64 {
-	if a < b {
-		return a
-	}
-	return b
-}
-
-func MaxF(a, b float64) float64 {
-	if a > b {
-		return a
-	}
-	return b
-}
-
 func StringReverse(s string) string {
 	r := []rune(s)
 	for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 {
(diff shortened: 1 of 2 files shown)


jedisct1:
For a given resolver, there's a high variance in the RTT according to whether a response was served from their cache or not, and their connectivity to authoritative servers. We can't select the "best" server without gradually learning its overall performance.

Resolvers that are slow or down can also become online and fast. This happens during temporary outages, overloads, but also when clients are switching VPN locations, or roaming on their mobile devices. Hence the idea of occasionally giving a chance to resolvers to go up the list.

jedisct1:
The current estimator sucks and is definitely worth improving though. The model is close to the [multi-armed bandit problem](https://en.wikipedia.org/wiki/Multi-armed_bandit) with the additional complexity that individual machines don't behave deterministically.

ghost:
I know. I don't want to re-invent the wheel, just make your code simpler and more effective.

E.g. to find the current best candidate you first choose a random one, switch it with the first element of the list and then bubble _up_ once, which might change the first element again, just to have a better, but not the best one.

I only bubble _down_ once to make the first element the current best candidate for sure. So why the additional work?

PR #2043, opened 2022-02-13
LBStrategy-aware estimator

ghost:
The estimator has at least 3 bugs:

1. When giving a new chance to a candidate the lowered RTTs converge to a value way above the best RTT in the list. This means that with most strategies most servers will never be tried again, i.e. one of the main features of the estimator is completely broken. 
2. Partial sorting isn't always applied when needed. However this might be rare enough to be ignored.
3. A new preferred candidate is sometimes found after it was already reported, i.e. the log might states a wrong one.

This commit only fixes bug 1.

I also adjusted RTTEwmaDecay to a more typical and stable value (α=0.1).

Note: Though it should now work as intended, it might decreases the user experience by trying out a slow server occasionally.

### DIFF
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -253,8 +253,7 @@ func (serversInfo *ServersInfo) estimatorUpdate() {
 	if activeCount == serversCount {
 		return
 	}
-	currentActive := rand.Intn(activeCount)
-	candidate := rand.Intn(serversCount-activeCount)+activeCount
+	candidate, currentActive := rand.Intn(serversCount-activeCount)+activeCount, rand.Intn(activeCount)
 	candidateRtt, currentActiveRtt := serversInfo.inner[candidate].rtt.Value(), serversInfo.inner[currentActive].rtt.Value()
 	if currentActiveRtt < 0 {
 		currentActiveRtt = candidateRtt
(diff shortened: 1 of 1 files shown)


jedisct1:
I fear this is going to make slow or unresponsive servers tried too much. Ideally, the penalty from a timeout should immediately move them back in the queue to avoid hurting too much the user experience. My intuition is that adding an instant response (`0.0`) rather than something close to existing best candidate is a bit too brutal.

ghost:
It's EWMA and 0.0 is added, not set. E.g. for a RTT of 100 we get the following sequence (rounded to integers):

(100, 90, 81, 73, 66, 59, 53, 48, 43, 39, 35, 31, 28, 25, 23, 21, 19, 17, 15, 14, etc.)

As soon as we fall below the current best RTT (e.g. 20) the estimator won't decrease anymore, so we land just below (19 in this example).

Or did I overlook something and there's a way for it to reach 0?

jedisct1:
Thanks for clarifying!

But can't the addition of `0.0` end up being applied to the current best candidate as well?

Don't we want to exclude the `n` (for a `pn` strategy) first entries from the promotion?

ghost:
The current best candidate is already protected. The rest is not, but it would make sense to do so.
I'll look into it tomorrow.

Now that your saying this, the estimator doesn't seem to be made for the current strategies but with something else in mind, strange.

jedisct1:
Looking really good! Thanks a lot for these tweaks!

Do you think this PR is ready to get merged?

ghost:
I guess so. Sorry for the noise at the end.

jedisct1:
Awesome, thank you!

PR #2045, opened 2022-02-16
restore old logic/constants

ghost:
Fixed 2 indices. The factor probably needs tuning or a completely new approach but I won't look into this anytime soon.


ghost:
Nope, still not there. Should get some sleep and think a few days about this.

ghost:
I ran some simulations and got the following results:

- If there is a restriction for lowering RTTs the estimator gets stuck on its own, only insanely high RTT fluctuations cause a reaction after this.
- Without a restriction inactive servers will be retried after some long intial lowering phase. After every inactive server has been tried at least once, retrying happens quickly and probably more often than one would like.
- Depending on the number of servers the intial lowering phase might need hundreds of queries before the first inactive is tried again. I guess that a certificate refresh is likely to happen earlier than this.

Because I didn't find any significant way of improvement I restore the old constants to ensure there's at least no degradation of user experience.

PR #2079, opened 2022-04-02
regression: fix ewma warmup again

ghost:
Though generally of advantage a warmup phase is detrimental to the lbStrategy estimator for several reasons. Therefore a new ewma variant with adjustable warmup phase was introduced in [7956ba5](https://github.com/DNSCrypt/dnscrypt-proxy/commit/7956ba5b104a1635214d2ddfbabb69b12ffefcf5) to fix this which worked fine.

Sadly a few commits later another ewma variant brought the warmup phase and (slightly different) issues back.

This commit reverts to the properly working ewma variant with adjustable warmup phase.

### DIFF
--- a/dnscrypt-proxy/estimators.go
+++ b/dnscrypt-proxy/estimators.go
@@ -3,7 +3,7 @@ package main
 import (
 	"sync"
 
-	"github.com/lifenjoiner/ewma"
+	"github.com/jedisct1/ewma"
 )
 
 const (
--- a/dnscrypt-proxy/estimators.go
+++ b/dnscrypt-proxy/estimators.go
@@ -13,7 +13,7 @@ const (
 type QuestionSizeEstimator struct {
 	sync.RWMutex
 	minQuestionSize int
-	ewma            *ewma.EWMA
+	ewma            ewma.MovingAverage
 }
 
 func NewQuestionSizeEstimator() QuestionSizeEstimator {
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -46,7 +46,7 @@ type DOHClientCreds struct {
 type ServerInfo struct {
 	DOHClientCreds     DOHClientCreds
 	lastActionTS       time.Time
-	rtt                *ewma.EWMA
+	rtt                ewma.MovingAverage
 	Name               string
 	HostName           string
 	UDPAddr            *net.UDPAddr
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -16,9 +16,9 @@ import (
 	"time"
 
 	"github.com/jedisct1/dlog"
+	"github.com/jedisct1/ewma"
 	clocksmith "github.com/jedisct1/go-clocksmith"
 	stamps "github.com/jedisct1/go-dnsstamps"
-	"github.com/lifenjoiner/ewma"
 	"github.com/miekg/dns"
 	"golang.org/x/crypto/ed25519"
 )
@@ -199,6 +199,7 @@ func (serversInfo *ServersInfo) refreshServer(proxy *Proxy, name string, stamp s
 		dlog.Fatalf("[%s] != [%s]", name, newServer.Name)
 	}
 	newServer.rtt = ewma.NewMovingAverage(RTTEwmaDecay)
+	newServer.rtt.SetWarmupSamples(1)
 	newServer.rtt.Set(float64(newServer.initialRtt))
 	isNew = true
 	serversInfo.Lock()
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/hectane/go-acl v0.0.0-20190604041725-da78bae5fc95
 	github.com/jedisct1/dlog v0.0.0-20210927135244-3381aa132e7f
+	github.com/jedisct1/ewma v1.2.1-0.20220220223311-a30af446ecb9
 	github.com/jedisct1/go-clocksmith v0.0.0-20210101121932-da382b963868
 	github.com/jedisct1/go-dnsstamps v0.0.0-20220328103132-6fb2da762370
 	github.com/jedisct1/go-hpke-compact v0.0.0-20210930135406-0763750339f0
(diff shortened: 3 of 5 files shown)


lifenjoiner:
I even thought there was no big problem with the original EWMA:
https://github.com/VividCortex/ewma/blob/487e8c9fe1e218b64696ee0219903a3a40d0dc2b/ewma.go#L120-L126
```go
// Set sets the EWMA's value.
func (e *VariableEWMA) Set(value float64) {
	e.value = value
	if e.count <= WARMUP_SAMPLES {
		e.count = WARMUP_SAMPLES + 1
	}
}
```
It is designed for the continuing case (restore and continue a previous calculation), and assuming the warmup has been finished.
So, there shouldn't be much difference between the original and our variants.

As I said in https://github.com/DNSCrypt/dnscrypt-proxy/issues/2046#issuecomment-1046431762, I still don't get it.
> Currently it needs 18 queries on average to get a meaningful answer for one server. However as only the first 2 servers are queried, one of the fastest ones now becomes the slowest, because the rest is still 0.
> So the intial good order would be scrambled and (if it was roughly correct) be brought back only after e.g. 180 queries on average assuming 10 live servers.

All available servers `rtt` have been initialized by `Set`, aren't they? How could they sill be 0?

lifenjoiner:
For my variant, use `Add` would be better here:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/77dc3b1e853950d354f505719df45b066fbea866/dnscrypt-proxy/serversInfo.go#L201-L202
And:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/77dc3b1e853950d354f505719df45b066fbea866/dnscrypt-proxy/estimators.go#L40

ghost:
Oh dear, sorry for screwing up my commit.

However using lifenjoiner's code is no option here, it's the worst variant of the 3 and causes a degradation of the estimator leaving us with a worse experience than the original implementation.

@lifenjoiner 
You are looking at the wrong part of the code. It's not about the initialization but the return value, see [here](https://github.com/VividCortex/ewma/blob/487e8c9fe1e218b64696ee0219903a3a40d0dc2b/ewma.go#L110-L118):
```
// Value returns the current value of the average, or 0.0 if the series hasn't
// warmed up yet.
func (e *VariableEWMA) Value() float64 {
	if e.count <= WARMUP_SAMPLES {
		return 0.0
	}

	return e.value
}
```
The estimator needs meaningful data to work with.

The original variant just returns 0.0 during the warmup phase. The estimator considered this valid data and shuffled around all the servers for no reason, even replacing fast servers with very slow ones for a long time. This was what I wanted to fix.

Your variant immediately returns measured values however they are computed differently than after the warmup phase. That screws up comparisons because those values are of the wrong scale and that often causes the estimator to make a wrong decision.

I'm only evaluating this under the perspective of the estimator: It doesn't matter if one of the variants would perform better after 100 measurements and would give a closer value to the actual mean. Because e.g. the estimator usually only gets the chance to compare an inactive server (measured once) to an active one (measured often) once or twice, so useful data is needed from the start and probably rarely again later. So except for the active servers the warmup phase behaviour is all that matters.

This all can easily be fixed by removing the warmup phase for the estimator (in other cases it's usually beneficial though).

lifenjoiner:
I was emphasizing `e.count = WARMUP_SAMPLES + 1`: the L123-L125.

ghost:
OK, but how does this matter here?

lifenjoiner:
Obviously, `e.count` > `WARMUP_SAMPLES`, `Value()` returns `e.value` rather than 0, and it will act like the warmup has been skipped.

lifenjoiner:
I'm not sure if you try my simulation in https://github.com/DNSCrypt/dnscrypt-proxy/issues/2046#issuecomment-1046431762. Let's show some data proofs:
ewma-cmp

ghost:
Just by looking at your plot:
What works best for the estimator is EWMA Continuing because it moves slowly without bigger jumps.
EWMA Warmup delivers no useful information at start but that's needed.
EMWA Variant jumps around to much at the beginning (outliers), that's again counterproductive.

Does this mean I mixed up the names and the original VividCortex was the best one all the time and I caused several useless commits and discussions? Then I'm very sorry.

Maybe someone else should take over here. However I don't want to be at fault for degrading the experience for all DNSCrypt users for probably many years. So please don't say it doesn't matter and make sure we end up with the gray line.

Sorry again to all for wasting your time.

@lifenjoiner 
Btw. how did you come up with your variant? It jumps around so much in the beginning, when would that be useful?

lifenjoiner:
First, I guess we should be on the same page: the validated servers ewma would not be 0. That's the keystone of your suspicion: the slowest becomes the fastest, and the fastest becomes the slowest.

2nd, I'm not saying my variant is the best. Compare and choose.
I've uploaded the simulation xlsx file with plot in https://github.com/lifenjoiner/ewma/issues/1, and also added explanations of the fields. Try it and observe the result you are interested in yourself.

Things I'll repeat here are:
1. The exact average (the blue line) is the ideal result. It should be the base that others should compare to and near to. Isn't it?
2. Adding the 1st sample by `Add` or `Set` is the key difference, which leads to "having warm-up or not" and "using just-in-time decay or not of my implement".
3. VividCortex's `VariableEWMA` and mine would be the same result, while both adding the 1st sample by `Set`.

lifenjoiner:
@livingentity
If you care about the `outliers`, that should be a work at an earlier stage, cleanup stage, but not in EWMA implementation. And then, you'd define `outliers` first.

ghost:
2. and 3. I only figured out yesterday, that was my problem. Thanks for taking the time to explain.

Taking care of outliers outside of EWMA would mean to store additional data. Within EWMA one could use e.g. simple ramp control.

I guess I still make a list of all the commits that would need to be reverted because of the garbage I proposed.
I should probably spent Frank and you a beer, too.

lifenjoiner:
I suppose we all want to make things better.

lifenjoiner:
More thoughts on the outliers at the "warmup" stage:
1. How can you tell which one is outlier, the 1st, the 2nd, or the 3rd? If none, exact average is better. If the 1st is, the average should be adjusted quicker/heavier. If one of the following is, the average should be adjusted slower/lighter.
2. What is the outlier ratio of a server? If it is high, the server is less reliable, adjusting quicker is better, the stable one will win at last. If all are low, they are reliable, exact average is better.

Both strategies are useful, it totally depends on how you use it.

Suspicion is not enough, we need proofs!

PR #2123, opened 2022-06-08
Reload expired client certificates

### LINKED ISSUE #2114: [DoH][TLS Client Authentication] Reload expired client certificates
## Output of the following commands:

```
# Others are less relevant here
➜  dnscrypt-proxy-macos_x86_64-2.1.1 ./dnscrypt-proxy -version
2.1.1
```

I have a DoH client server which only uses short-lived client certificates (2 week validity) when connecting to a mutually authenticated DoH server. These certificates are renewed weekly, replacing the existing file(s). However, since `dnscrypt-proxy` does not refresh its client credentials, all DoH requests fail due to TLS errors once the client certificate has expired, provided the upstream is enforcing valid client certificates. This means that my client only has between 1-2 weeks of use before I need to restart the `dnscrypt-proxy` process to pick up the new certificate.

## *When* does this occur?
Upon startup, `dnscrypt-proxy` loads the client certificate into memory when building the `xtransport`. However, they are never updated after that. Should the certificate in memory expire, all requests which use it will fail. There are a few places where the transport is rebuilt, but none trigger on this particular error (only on "handshake failure").

## *Where* does it happen?
It happens when a DoH request is being performed while the client certificate loaded into memory has expired.
The relevant error handling will likely take place here:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/c367a82ac069668a36b5478ef2cf36478f664033/dnscrypt-proxy/xtransport.go#L437-L447

1. Generate a short-lived certificate (as short as a few minutes, if you'd like)
2. Start `dnscrypt-proxy` with this client certificate, targeting a DoH server which supports mutual authentication and enforces certificate validity
3. Generate a replacement certificate, with a later expiration date
4. Wait until the original certificate has expired
5. All requests should now fail, with an error similar to the following found in the (verbose) logs:
```
[2022-05-13 16:47:19] [DEBUG] [https://<redacted>/dns-query?dns=<redacted>]: [Get "https://<redacted>/dns-query?dns=<redacted>": remote error: tls: bad certificate]
```

There are likely (at least) two ways to go about solving this issue:
1. (Harder) Monitor the certificate files, reloading them when they update
2. (Easier) Rebuild the transport upon "bad certificate", similar to what is done today when "handshake failure" is encountered

## Other Comments
As a stopgap, I have updated my DoH endpoints to accept expired client certificates for now, but this is obviously not the ideal end state here.

mjdiffy:
When the client credentials used for mutual authentication expire, all requests
to a DoH server which enforces validitiy will fail. To mitigate this, we can
try to reload the existing credentials from disk. This assumes that the
credentials have been rotated outside of the dnscrypt-proxy process and thus a
reload will pick up the newest versions.

The simplest way to do this is to simply rebuild the transport upon receiving
an error which contains the string "bad certificate".

Resolves #2114

### DIFF
--- a/dnscrypt-proxy/xtransport.go
+++ b/dnscrypt-proxy/xtransport.go
@@ -442,6 +442,11 @@ func (xTransport *XTransport) Fetch(
 			)
 			xTransport.tlsCipherSuite = nil
 			xTransport.rebuildTransport()
+		} else if len(xTransport.tlsClientCreds.clientCert) > 0 && strings.Contains(err.Error(), "bad certificate") {
+			dlog.Warnf(
+				"TLS certificate failure - Check the validity of the client certificate; reloading existing certificate",
+			)
+			xTransport.rebuildTransport()
 		}
 		return nil, statusCode, nil, rtt, err
 	}


mjdiffy:
To test my changes, I no longer had a mutually-authenticated DoH server which _does_ enforce certificate validity, so I pointed it to another random HTTPS endpoint which did. However, it doesn't speak DoH, so I didn't expect valid outputs when it got there:
```
➜  dnscrypt-proxy git:(master) ✗ go run . --config ~/dnscrypt-proxy.toml
[2022-06-08 15:27:37] [NOTICE] dnscrypt-proxy 2.1.1
[2022-06-08 15:27:37] [NOTICE] Enabling TLS authentication
Reload expired client certificates
[2022-06-08 15:27:37] [NOTICE] Now listening to 127.0.0.1:53443 [UDP]
[2022-06-08 15:27:37] [NOTICE] Now listening to 127.0.0.1:53443 [TCP]
[2022-06-08 15:27:37] [NOTICE] Firefox workaround initialized
[2022-06-08 15:27:38] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:38] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:38] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:38] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:38] [ERROR] Get "https://<redacted>": remote error: tls: bad certificate
[2022-06-08 15:27:38] [NOTICE] dnscrypt-proxy is waiting for at least one server to be reachable
[2022-06-08 15:27:48] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:48] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:48] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:48] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
[2022-06-08 15:27:58] [WARNING] TLS certificate failure - Check the validity of the client certificate; reloading existing certificate
```
At this point, I flipped the certificate file to a non-expired version, and it reloaded properly, but now I had the issue of pointing at a non-DoH server, but at least what I was after worked!
```
[2022-06-08 15:27:58] [WARNING] [dohproxy_v6]: dns: bad rdata
[2022-06-08 15:27:59] [WARNING] [dohproxy_v4]: dns: bad rdata
[2022-06-08 15:28:09] [WARNING] [dohproxy_v6]: dns: bad rdata
[2022-06-08 15:28:09] [WARNING] [dohproxy_v4]: dns: bad rdata
# ...
```

jedisct1:
Thanks!

It may work as a quick hack, but this is not something that can be merged as-is.

If a single server has a permanent certificate error, that would cause the entire configuration to be reloaded every time it gets a query. Not great if more than one server is in use, and that can even be considered a remotely exploitable security issue.

jedisct1:
Also, the same issue may affect ODoH relays.

PR #2124, opened 2022-06-11
PluginDNS64: Use read and write mutexes as approp

ignoramous:
If the lock is intended to protect `plugin.pref64`, then the current change may make sense. Otherwise, I might have gotten this commit upside down.


jedisct1:
I'm not too familiar with that part of the code, but indeed, the lock types look reversed.

jedisct1:
Merged. @s-s let us know if that doesn't look right to you :)

PR #2125, opened 2022-06-13
Use registeredServers slice copy during ServerInfo refreshing period

lifenjoiner:
goroutines:
[proxy.updateRegisteredServers()](https://github.com/DNSCrypt/dnscrypt-proxy/blob/8d737a69f550158ed6a08aef46bdb975ac6e47cb/dnscrypt-proxy/proxy.go#L257) versus [proxy.serversInfo.refresh(proxy)](https://github.com/DNSCrypt/dnscrypt-proxy/blob/8d737a69f550158ed6a08aef46bdb975ac6e47cb/dnscrypt-proxy/proxy.go#L269)


lifenjoiner:
The type of `serversInfo.registeredServers` is slice.
`=` only copies the pointer of slice (https://gist.github.com/lifenjoiner/dc0ac1c933ff836094c1c6b123902714), which is different from array. It's really a "trick" of go. 😂
The refreshing takes a while.

jedisct1:
Woah, this is so confusing.

PR #2127, opened 2022-06-14
RUnlock for early exit

lifenjoiner:


jedisct1:
Thank you so much for reviewing all these locking issues!

lifenjoiner:
This is a very useful tool. I also learn a lot here.

PR #2130, opened 2022-06-15
Unlock before return

syphyr:
Add missing unlock before return


jedisct1:
This is a lambda function, we don't return from the actual function.

PR #2136, opened 2022-06-28
Fix goroutines memory leak by unbuffered channel blocking

lifenjoiner:
Seems there is a mysterious issue of memory climbing up, uncaught, I suspected.
That is because of, in my mind, `dnscrypt-proxy` will cost 30+M private memory after several days, while it costs ~13M at launch.

Recently, I did some profiling and get the following details.

1. A `typeperf` (on Windows) samples analysis. They are sampled per 5 minutes, and 573 samples.
    2865m

    `dnscrypt-proxy`: upstream DNS of `dnsforwarder`, no caching, no any lists.
    `dnsforwarder`: default DNS, caching, types of lists.
    `pd`: http proxy, important traffic indicator, also a go program.

    Here we can see the climbing up. That will be strange, considering I use it as a bare proxy. The comparison of the 3 tells it is unlikely the system memory management issue.

2. I did `go tool pprof` on `dnscrypt-proxy`, and found it is the goroutines, rather than the heap.
    There were about 160 goroutines while it is idle, after working 7h. Too much! Built on https://github.com/DNSCrypt/dnscrypt-proxy/commit/59ce17e0abe8297182f19ed2e841051193f62ad3.
    ```
    goroutine profile: total 163
    82 @ 0xc3a48f 0xc04a40 0xc04734 0xf86815 0xc666b1
    #	0xf86814	main.DNSExchange.func1+0xa4	github.com/dnscrypt/dnscrypt-proxy/dnscrypt-proxy/dnsutils.go:331

    71 @ 0xc3a48f 0xc04a40 0xc04734 0xf866d5 0xc666b1
    #	0xf866d4	main.DNSExchange.func2+0xa4	github.com/dnscrypt/dnscrypt-proxy/dnscrypt-proxy/dnsutils.go:347
    ```
    The channel blocks, and the goroutine hangs on!

Blocking PoC for **unbuffered channels** Vs **buffered channels**: https://gist.github.com/lifenjoiner/16d279ddfc0e6b81b1d4082c6cff3c1f

With this patch, the goroutines total will be about 14 while it is idle, after working 7h.

Here is a comparison for the first 8h:
Before:

After:

The positive correlation with `pd` can be seen.

I'm still profiling the memory usage for longer term performance ... Hope there will be good news.

### DIFF
--- a/dnscrypt-proxy/coldstart.go
+++ b/dnscrypt-proxy/coldstart.go
@@ -84,6 +84,7 @@ func handleColdStartClient(clientPc *net.UDPConn, cancelChannel chan struct{}, i
 	exit := false
 	select {
 	case <-cancelChannel:
+		cancelChannel <- struct{}{}
 		exit = true
 	default:
 	}
@@ -95,6 +96,12 @@ func handleColdStartClient(clientPc *net.UDPConn, cancelChannel chan struct{}, i
 	}
 	if err != nil {
 		dlog.Warn(err)
+		select {
+		case <-cancelChannel:
+			cancelChannel <- struct{}{}
+		default:
+			cancelChannel = make(chan struct{}, 1)
+		}
 		return true
 	}
 	packet := buffer[:length]
@@ -134,7 +141,6 @@ func addColdStartListener(
 		for !handleColdStartClient(clientPc, cancelChannel, ipsMap) {
 		}
 		clientPc.Close()
-		cancelChannel <- struct{}{}
 	}()
 	return nil
 }
--- a/dnscrypt-proxy/dnsutils.go
+++ b/dnscrypt-proxy/dnsutils.go
@@ -316,7 +316,7 @@ func DNSExchange(
 ) (*dns.Msg, time.Duration, bool, error) {
 	for {
 		cancelChannel := make(chan struct{})
-		channel := make(chan DNSExchangeResponse)
+		channel := make(chan DNSExchangeResponse, 6)
 		var err error
 		options := 0
 
@@ -325,32 +325,32 @@ func DNSExchange(
 				queryCopy := query.Copy()
 				queryCopy.Id += uint16(options)
 				go func(query *dns.Msg, delay time.Duration) {
-					option := _dnsExchange(proxy, proto, query, serverAddress, relay, 1500)
+					time.Sleep(delay)
+					option := DNSExchangeResponse{err: errors.New("Canceled")}
+					select {
+					case <-cancelChannel:
+					default:
+						option = _dnsExchange(proxy, proto, query, serverAddress, relay, 1500)
+					}
 					option.fragmentsBlocked = false
 					option.priority = 0
 					channel <- option
-					time.Sleep(delay)
-					select {
-					case <-cancelChannel:
-						return
-					default:
-					}
 				}(queryCopy, time.Duration(200*tries)*time.Millisecond)
 				options++
 			}
 			queryCopy := query.Copy()
 			queryCopy.Id += uint16(options)
 			go func(query *dns.Msg, delay time.Duration) {
-				option := _dnsExchange(proxy, proto, query, serverAddress, relay, 480)
+				time.Sleep(delay)
+				option := DNSExchangeResponse{err: errors.New("Canceled")}
+				select {
+				case <-cancelChannel:
+				default:
+					option = _dnsExchange(proxy, proto, query, serverAddress, relay, 480)
+				}
 				option.fragmentsBlocked = true
 				option.priority = 1
 				channel <- option
-				time.Sleep(delay)
-				select {
-				case <-cancelChannel:
-					return
-				default:
-				}
 			}(queryCopy, time.Duration(250*tries)*time.Millisecond)
 			options++
 		}


dnscrypt-proxy/dnsutils.go
```diff
@@ -316,7 +316,7 @@ func DNSExchange(
 ) (*dns.Msg, time.Duration, bool, error) {
 	for {
 		cancelChannel := make(chan struct{})
-		channel := make(chan DNSExchangeResponse)
+		channel := make(chan DNSExchangeResponse, 6)
```
jedisct1, line 319: Where does this number comes from?

dnscrypt-proxy/dnsutils.go
```diff
@@ -325,32 +325,32 @@ func DNSExchange(
...
 				go func(query *dns.Msg, delay time.Duration) {
-					option := _dnsExchange(proxy, proto, query, serverAddress, relay, 1500)
-					option.fragmentsBlocked = false
-					option.priority = 0
-					channel <- option
 					time.Sleep(delay)
+					option := DNSExchangeResponse{err: errors.New("Canceled")}
 					select {
 					case <-cancelChannel:
-						return
```
jedisct1, line 335: Shouldn't we still return here?

jedisct1:
Wow, this is great findings.

I must confess that I'm still a little bit confused about the difference between unbuffered and buffered channels. Don't buffered channels block if they are empty as well?

Anyway, these changes look great! Having all these dangling goroutines was definitely not intentional.

dnscrypt-proxy/dnsutils.go
lifenjoiner: Use 2*MaxTries to set the buffered channel size.

dnscrypt-proxy/dnsutils.go
lifenjoiner: I don't know if it is possible `cancelChannel` is corrupted before use, or closed unintentionally in the future, just felt unsafe if yield nothing to the channel. Sending a dummy record is low cost.
Of course, the possibility is very low. You can optimize it.

lifenjoiner:
> Don't buffered channels block if they are empty as well?

Whether sending to a channel blocks, depends on whether there is room, not whether the channel is empty. If the capacity is neither 0 nor fulfilled, sending doesn't block.
Receiving from empty channel blocks.
You can run this [demo](https://gist.github.com/lifenjoiner/16d279ddfc0e6b81b1d4082c6cff3c1f), and see sending to channel b is non-blocking.

lifenjoiner:
Make a summary:
Whether sending or receiving blocks, depends on whether there is resource available. :)

alisonatwork:
@jedisct1 is there a chance to release a new version for this fix? 2.1.1 is quite old now.

jedisct1:
@alisonatwork Yes, this is indeed a good opportunity to tag a new release. I'll do it shortly.

lifenjoiner:
Good news:
I think it is stable according my tests on 2 PCs lasting 5 days.
1534-128h
1534 samples. Blank range is hibernation.
The other one with daily hibernation is also stable.
Their peeks didn't up to 20M.
Without pprof module would be better. 😄

dnscrypt-proxy/coldstart.go
```diff
@@ -95,6 +96,12 @@ func handleColdStartClient(clientPc *net.UDPConn, cancelChannel chan struct{}, i
 	}
 	if err != nil {
 		dlog.Warn(err)
+		select {
+		case <-cancelChannel:
+			cancelChannel <- struct{}{}
+		default:
+			cancelChannel = make(chan struct{}, 1)
```
jedisct1, line 103: Is that line correct? We create a channel, but we immediately return so nothing's gonna be pushed to it.

jedisct1: (yes, that's a late comment, but I stumbled upon that right before tagging a release)

lifenjoiner: It is for these unbuffered channels:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/3d641b758a98f712ed95e4ffc89eff327c139679/dnscrypt-proxy/coldstart.go#L21-L26
The workflow is:
1. `Stop` tries to send a signal to each `handleColdStartClient` to exit. It blocks until a `select` in `handleColdStartClient` receives from the channel and then delivers a new value to equal the send/receive balance.
2. The 2 `return true` of `handleColdStartClient` will end rerunning `handleColdStartClient`. The 2nd `select` is in case of the `Stop` sends signals after the 1st one. If that is between the 2 `select`s, we relieve the channel blocking. If signal is still not sent, we change the channel to a buffered channel, which doesn't block. The 1st `return true` is from the 1st `select`, and has had a balanced consumed channel.

A [demo](https://gist.github.com/lifenjoiner/6de299af7b8ef5a27ac6c8ea1160c3e4) may help on this unbuffered channel and select behaviors topic.

lifenjoiner: I made it complicated. I understand it better now.
A buffered channel is good enough. It is non-blocking for sending, for `default` case of `select`, and for `select` on closed channel too.
Optimized with https://github.com/DNSCrypt/dnscrypt-proxy/pull/2151.

lifenjoiner: Leave a channel open is OK: https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open, but blocked is not a good choice.

jedisct1: Yay, thank you so much for this!

PR #2148, opened 2022-07-08
Reduce comparisons

lifenjoiner:


jedisct1:
Probably not something that is making a significant difference, but sure!

Thank you!

lifenjoiner:
Minor. This is a mature program!

PR #2159, opened 2022-07-22
xtransport layer to netip and immediate dependencies

ianbashford:
@jedisct1 
First step towards converting dnscrypt to use netip.Addr in place of net.IP.   I'm hoping we'll see some efficiency gains as the objects are tighter.   The original article is here https://tailscale.com/blog/netaddr-new-ip-type-for-go/  and has now been adopted into the 'netip' package.
I've just tackled the xtransport layer for now - if you're happy with the swap, then I'll start on the plugins and other areas.
I don't see signs of miekg updating the dns package to netip, but maybe the V2 he's working on might... 

One specific thing to bring your attention to in `xtransport.go`, lines `276` and `296`.  This is the `miekg/dns -> dnscrypt` interface:
The original code assumes, that if there are answers, then the 'IP' returned in the answer will be used (no nil checks).
I've continued with that assumption, and don't check for an err object parsing the result; happy to discuss.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -926,7 +926,7 @@ func cdLocal() {
 
 func isIPAndPort(addrStr string) error {
 	host, port := ExtractHostAndPort(addrStr, -1)
-	if ip := ParseIP(host); ip == nil {
+	if _, err := ParseIP(host); err != nil {
 		return fmt.Errorf("Host does not parse as IP '%s'", addrStr)
 	} else if port == -1 {
 		return fmt.Errorf("Port missing '%s'", addrStr)
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -260,7 +260,7 @@ func (serversInfo *ServersInfo) estimatorUpdate(currentActive int) {
 	if activeCount == serversCount {
 		return
 	}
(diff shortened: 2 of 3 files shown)


jedisct1:
Hi Ian!

And sorry for the delayed response.

Moving to `netip` is a good thing, thank you for kicking this off!

I'll merge this after version 2.1.2 is released.

ianbashford:
Hi Frank
Good news.  In that case I'll start working my way around the plugins and anything else next...

kfken:
There's a couple errors in this commit related to use of netip that I think need to be addressed:

1) The new `xtransport.ParseIP()` doesn't match the intended usage.  In the original code, it retuned nil if the host wasn't an IP.  `netip.ParseAddr()` returns `err == nil` when host is an IP.  So by just swapping `ip` with `err`, in the comparisons, the logic is swapped and the wrong code path is taken.  

2) Using `netip.IPv4Unspecified()` doesn't match the original intent of the code that was replaced.  `IPv4Unspecified()` returns 0.0.0.0, but the original code was checking if (for example) `cachedIP == nil`.  `net.IP` is a `[]byte`, and the zero value is `nil`.  So the new zero value that should be used from netip is `netip.Addr{}`.   

3. Also, using `IPv4Unspecified()` completely misses IPv6 resolvers (yes they really are used).  There's also usage of `netip.Is4()` which similarly breaks support for v6 addresses.

PR #2167, opened 2022-08-03
Log to console when in command mode

lifenjoiner:
For https://github.com/DNSCrypt/dnscrypt-proxy/discussions/2166.
Quick results.
Avoid overwriting the log file in use, by the same config most of the time.


jedisct1:
Looks good!

PR #2188, opened 2022-08-25
Optimize adopted relay name to show

lifenjoiner:
DNSCrypt relay requires ServerAddrStr;
ODoH relay requires ProviderName, port 443 can be either present or not;
raw stamp can be both.

Displaying specified stamp makes it easier to debug.


jedisct1:
Thank you!

PR #2204, opened 2022-09-21
Try dnscrypt-proxy to resolve configured hosts when ignore_system_dns

lifenjoiner:
Trying to relieve the concern like https://github.com/DNSCrypt/dnscrypt-proxy/discussions/2201.

### DIFF
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -258,7 +258,15 @@ cert_refresh_delay = 240
 bootstrap_resolvers = ['9.9.9.11:53', '8.8.8.8:53']
 
 
-## Always use the bootstrap resolver before the system DNS settings.
+## This is a switch for prior bootstrap resolvers.
+##
+## Most of the time, the system DNS leads to dnscrypt-proxy itself,
+## if you set up dnscrypt-proxy in the system DNS chain.
+##
+## If true, dnscrypt-proxy will directly try on its own service instead.
+## Together with bootstrap resolvers, upstream hosts will be resolved most
+## of the time.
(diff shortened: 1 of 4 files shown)


lifenjoiner:
Generally to users, `dnscrypt-proxy` has been set to their system DNS, use-system-DNS goes to:
1. Most of the time, `resolveUsingResolvers` queries are redirected back to `dnscrypt-proxy` service.
    There could be a `NOT_READY` delay at startup.
2. If `dnscrypt-proxy` is not in the system DNS or there are other DNS when `dnscrypt-proxy` is not usable (try first then know), the queries are redirected to the specified DNS.

With an internal query, we can skip the `NOT_READY` period, and switch back ASAP.
And `dnscrypt-proxy` works well when there is a service AFAIK. Querying to itself will obtain more privacy rather than the bootstrap resolvers, although that is not a big problem some may concern.
So, we can have an internal try before querying the normal DNS.

System DNS forks and making `ignore_system_dns` a "hybrid". I'm wondering if we can remove it. If users prefer a local DNS they can specify it in the `bootstrap_resolvers` explicitly. That will make things more clear. But it would be a breaking change to the config.

Any comments?

jedisct1:
Sorry for the delay.

That sounds like a good idea. Why did you close this?

lifenjoiner:
I was confused if this is an expected direction. I'll reopen it, as you say it sounds like a good idea.

Append a scenario which may need "use system DNS"/`ignore_system_dns`:
`dnscrypt-proxy` runs on a router as a LAN DNS server, and it is not set as the router's DNS. Using system DNS would be a lazy but convenient option for those who prefer ISP's DNS as prior bootstrap resolvers. This would be a reason to keep the option.

lifenjoiner:
Still think it is an idea or feature that can be discussed.

PR #2331, opened 2023-02-20
`dlog.SetLogLevel(dlog.SeverityDebug)` if `go test -v`

lifenjoiner:


jedisct1:
Good idea, thank you!

PR #2334, opened 2023-02-23
Adopt `refresh_delay`s as cache sources TTL

lifenjoiner:
`refresh_delay` was limited to `[1, 168]`, while `source.prefetchDelay` is 24 which controlled downloading.

That makes `refresh_delay` less effective.
`[1, 24)` triggers reparsing cache sources but no downloading, `(24, 168]` is defeated by `source.prefetchDelay` but says still fresh. In other words, value which is non-integer times of 24 is ceiled, and downloading always happen every 24h if there is no failure triggers retry.

The process flow was complicated and puzzling. Plus there could be multiple `refresh_delay`s.

Now, both downloading and reparsing are ruled by `refresh_delay`, `prefetchDelay` is obsoleted. It gives user sense of mastery, with tutelage.

---
More explains:

The delay algorithm of source downloading and reparsing is very complicated and puzzling.
Here are my understandings:
Workflow
delay
Delay
sleep

Different `refresh_delay`s of sources make things more complicated. Here are some key rules:
1. `prefetchDelay` always controls the downloading, if no failure. At startup, `refresh_delay` doesn’t determine delay at all. If the cache expired, then the delay depends on the downloading result. If the cache isn’t expired, ` delay = source.prefetchDelay - elapsed`.
2. `refresh_delay` is only effective at startup or on out-of-step sources, when it `< prefetchDelay` and the cache is expired at startup. When it `> prefetchDelay`, `source.prefetchDelay - elapsed` is smaller. So, if the cache elapsed between them, program will say the cache is fresh but still is going to download the files, and then sleep depending on the result. `refresh_delay` doesn’t produce the delay too, when the flow starts from `PrefetchSources`.
3. Any cache delay remain time is ceiled to prefetchDelay. Those `refresh_delay > prefetchDelay`, but is not `N * prefetchDelay`, the `refresh_delay % prefetchDelay` will be ceiled to prefetchDelay. And `refresh_delay < prefetchDelay` too. It makes `refresh_delay` slightly meaningless. And, at startup, the worst case delay is approximately `2 * prefetchDelay`.

In summary, `refresh_delay` is less effective and makes things puzzling. Adopting `refresh_delay` as the real delay will make things simple and give user sense of mastery.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -878,8 +878,8 @@ func (config *Config) loadSource(proxy *Proxy, cfgSourceName string, cfgSource *
 	if cfgSource.FormatStr == "" {
 		cfgSource.FormatStr = "v2"
 	}
-	if cfgSource.RefreshDelay <= 0 {
-		cfgSource.RefreshDelay = 72
+	if cfgSource.RefreshDelay < 24 {
+		cfgSource.RefreshDelay = 24
 	} else if cfgSource.RefreshDelay > 168 {
 		cfgSource.RefreshDelay = 168
 	}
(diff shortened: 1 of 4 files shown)


nanis94:
master

nanis94:
master

PR #2337, opened 2023-02-24
Fix DoH3 connections piling up

lifenjoiner:
DoH3 creates a new connection for each request without closing.

* `Conn` should be self maintained if it's created by customized `Dial` of `http3.RoundTripper`. https://pkg.go.dev/github.com/quic-go/quic-go#DialAddrEarlyContext

* http3 doesn't have a `CloseIdleConnections`. https://pkg.go.dev/net/http#Client.CloseIdleConnections


jedisct1:
Good catch!

PR #2364, opened 2023-04-06
Update the minisig file timestamp together with the source file

lifenjoiner:


jedisct1:
Sorry, I don't understand that code at all.

What does `writeToCache()` do? How is it different from `writeSource()`? Why do timestamps have to be manually changed?

I'm completely lost.

lifenjoiner:
Whaaat? ChatGPT?
If `bytes.Equal(source.in, bin)`, the files won't be touched. And, `fetchWithCache` requires the timestamp of the source file.

lifenjoiner:
Seems it has been patched.

jedisct1:
I didn't remember we were doing such a thing. Cool.

The whole way we handle sources is weird, though. Looks like we are keeping a copy of the markdown files in memory to parse them later. I don't get why the files are not parsed at the same time as they are read.

PR #2367, opened 2023-04-07
Add DOH certificate generation commands into the example config.

KOLANICH:


jedisct1:
Thank you!

syphyr:
Is there a reason why the commands to generate localhost.pem changed from what the wiki currently says?

openssl req -x509 -nodes -newkey rsa:2048 -days 5000 -sha256 -keyout \
  localhost.pem -out localhost.pem

KOLANICH:
RSA is a heavy asymmetric crypto method; since this crypto is not really used for security (we use a resolver under our control), I have changed it to a lightweight ECC suitable even for embedded devices (it is used in OpenWRT by default, if TLS is enabled in Web GUI). Though this curve is not in the [list of safe curves](https://safecurves.cr.yp.to/), again, we use encryption in that case only to workaround the flaws in Firefox that seem will never be fixed.

jedisct1:
RSA verification is faster than ECC.

And according to the recent activity on the relevant ticket, looks like Firefox will eventually fix the ECH issue.

KOLANICH:
Thank you for the info.

PR #2421, opened 2023-06-24
Fix miscellaneous style issues

c4rlo:
Found by running `golangci-lint run --enable-all` (except the `errcheck` issues are dealt with separately in #2420).

I have only addressed the reported issues that seemed relevant to me.


jedisct1:
Thanks!

This is mostly for paranoia, but I'd rather keep checking for negative length, as it's not more expensive than an equality.

PR #2426, opened 2023-07-02
Balance multiple stamps with the same name

lifenjoiner:
It used to replace the old one with the 1st new found. That is rotating between the first two.
https://github.com/DNSCrypt/dnscrypt-proxy/discussions/2422

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -297,6 +297,10 @@ func (proxy *Proxy) updateRegisteredServers() error {
 				len(registeredServers),
 			)
 		}
+		// Balance multiple stamps with the same name
+		rand.Shuffle(len(registeredServers), func(i, j int) {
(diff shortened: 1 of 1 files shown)


lifenjoiner:
[Entries with 3+ stamps](https://github.com/DNSCrypt/dnscrypt-resolvers/blob/5b3eed5d90832b2f51a7b01e580edc31beb02d78/v3/public-resolvers.md?plain=1#L1781-L1787) don't work well. The 3rd and following stamps won't be adopted. Do they? Or do I missed anything?
Unfortunately, those servers are unavailable on my side to test this.

lifenjoiner:
Shuffled at:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/16b2c84147737315ae6a82bb62c3e4a89cd701a2/dnscrypt-proxy/sources.go#L295-L297

PR #2438, opened 2023-07-15
Add .mail & .home.arpa undelegated names

Expertcoderz:
Both names have been recognized for internal use in private networks.

`mail.` — Delegation rejected by [ICANN resolution 2018.02.04.12](https://www.icann.org/en/board-activities-and-meetings/materials/approved-board-resolutions-regular-meeting-of-the-icann-board-04-02-2018-en#2.c) (alongside `corp.` & `home.`) owing to its pertinent use in internal networks.

`home.arpa.` — Formally designated by [RFC 8375](https://datatracker.ietf.org/doc/html/rfc8375) as a special-use domain name for use within residential networks; supplanting `home.` for HNCP as well. According to the RFC:

>  DNS queries for names ending with '.home.arpa.' are resolved using
   local resolvers on the homenet.  Such queries MUST NOT be recursively
   forwarded to servers outside the logical boundaries of the homenet.


jedisct1:
Thank you!

PR #2457, opened 2023-08-11
Configured Github CI to always use latest go version

Neo2308:
* Check if there is a newer version of go before using the cached version.


jedisct1:
Thank you!

PR #2470, opened 2023-08-25
Harmonize `block_ipv6` with `blocked_query_response`

lifenjoiner:


jedisct1:
There's an important difference between IPv6 blocks and other blocks.

With other blocks, if they cause clients to block all record types of the same name, this is fine in practice.

With IPv6 blocks, only that specific record type should be interpreted as nonexistent.

At the beginning, IPv6 blocking was handled the same way as other blocks.

That caused issues with clients doing happy eyeball, with email gateways and with caches. That also caused issues when it was the target of a CNAME record.

The current solution is what causes the minimum amount of incompatibilities.

The message is also useful. With other blocks, the reason for the block is visible in logs. So, when an expected response is not received, it's possible to understand why by looking at the logs.

With IPv6 blocking, nothing is logged. Which is good because otherwise it would be very noisy (for every query) and redundant.

Having the message is thus helpful to understand what's going on. The message that suggests reviewing the `reject_ipv6` setting was added because users complained that domains didn't resolve, and the relationship with  the `reject_ipv6` option in their configuration file was not obvious.

lifenjoiner:
Thanks for explaining! And maybe descriptions in the toml file could be improved, to avoid leaving an impression that it does not keep the promise. I did not see any essential difference from the types of empty answers, but preference of OSes.

Today I Learned: Happy Eyeballs, a better strategy when dual-stack connectivities are available.

PR #2537, opened 2023-11-23
Support server refresh concurrency

keatonLiu:
We can simultaneously refresh all servers instead of refresh them one by one, which can greatly improve startup speed without visible loss of accuracy

### DIFF
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -229,12 +229,23 @@ func (serversInfo *ServersInfo) refresh(proxy *Proxy) (int, error) {
 	serversInfo.RUnlock()
 	liveServers := 0
 	var err error
-	for _, registeredServer := range registeredServers {
-		if err = serversInfo.refreshServer(proxy, registeredServer.name, registeredServer.stamp); err == nil {
-			liveServers++
-			proxy.xTransport.internalResolverReady = true
-		}
+
+	// simultaneously refresh all servers
+	wg := sync.WaitGroup{}
+	wg.Add(len(registeredServers))
+	for i := range registeredServers {
+		go func(rs *RegisteredServer) {
+			if err = serversInfo.refreshServer(proxy, rs.name, rs.stamp); err == nil {
+				serversInfo.Lock()
+				liveServers++
+				proxy.xTransport.internalResolverReady = true
+				serversInfo.Unlock()
+			}
+			wg.Done()
+		}(&registeredServers[i])
 	}
+	wg.Wait()
+
 	serversInfo.Lock()
 	sort.SliceStable(serversInfo.inner, func(i, j int) bool {
 		return serversInfo.inner[i].initialRtt < serversInfo.inner[j].initialRtt
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -224,27 +224,26 @@ func (serversInfo *ServersInfo) refresh(proxy *Proxy) (int, error) {
 	dlog.Debug("Refreshing certificates")
 	serversInfo.RLock()
 	// Appending registeredServers slice from sources may allocate new memory.
-	registeredServers := make([]RegisteredServer, len(serversInfo.registeredServers))
+	serversCount := len(serversInfo.registeredServers)
+	registeredServers := make([]RegisteredServer, serversCount)
 	copy(registeredServers, serversInfo.registeredServers)
 	serversInfo.RUnlock()
-
 	countChannel := make(chan struct{}, proxy.certRefreshConcurrency)
-	waitChannel := make(chan struct{})
-	var err error
+	errorChannel := make(chan error, serversCount)
 	for i := range registeredServers {
 		countChannel <- struct{}{}
 		go func(registeredServer *RegisteredServer) {
-			if err = serversInfo.refreshServer(proxy, registeredServer.name, registeredServer.stamp); err == nil {
+			err := serversInfo.refreshServer(proxy, registeredServer.name, registeredServer.stamp)
+			if err == nil {
 				proxy.xTransport.internalResolverReady = true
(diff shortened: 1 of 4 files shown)


jedisct1:
Sure, but by doing so, the latency measurements become unreliable. Not so much of an issue with a super snappy WiFi network, but this is an issue when the link is not that great.

jedisct1:
For DoH/ODoH/DoOH servers without an IP address, most of these would also use the fallback servers, which is something we want to minimize. With sequential refreshes, as soon as a server is ready, fallback servers can be ignored.

jedisct1:
I'm also a bit concerned about CPU/memory usage, as a lot of people are running `dnscrypt-proxy` on routers with limited resources. It may cause the server to not be able to start at all, or to freeze due to CPU spikes. 

Maybe that should be configurable.

keatonLiu:
I agree that it can be configurable, because for most home PC, the bandwidth and cpu cost can be ignored, while achieving significant performance improvement. If use sequential refresh, it may take several minutes to find an optimal path, compared to only a few seconds for simultaneous refresh

lifenjoiner:
The model is: burst at a time, finish in several seconds.

I would suggest setting the Maximum Concurrency.

Besides the advantage and previous disadvantages, think about:
1. 300+ at a time could be suspicious, in security.
2. If there is a network accident for 10s, there could be no available servers.

lifenjoiner:
@keatonLiu 
It's a pity that a good idea lost contact.
I've pushed my [suggestions](https://github.com/keatonLiu/dnscrypt-proxy/pull/1) to your [update](https://github.com/keatonLiu/dnscrypt-proxy/tree/update) branch.

keatonLiu:
> @keatonLiu It's a pity that a good idea lost contact. I've pushed my [suggestions](https://github.com/keatonLiu/dnscrypt-proxy/pull/1) to your [update](https://github.com/keatonLiu/dnscrypt-proxy/tree/update) branch.

Thank you for your ideas, I'm recently busy looking for an internship, I'll check it out and update soon.

keatonLiu:
I think if the liveServers variable is not atomic there will be a race: 6d179bbec8533928516735dab8749af60813ad2c

lifenjoiner:
Return `innerLen` and remove `liveServers` is simpler.

keatonLiu:
> Return `innerLen` and remove `liveServers` is simpler.

Sure!

dnscrypt-proxy/serversInfo.go
```diff
@@ -227,13 +227,26 @@ func (serversInfo *ServersInfo) refresh(proxy *Proxy) (int, error) {
 	registeredServers := make([]RegisteredServer, len(serversInfo.registeredServers))
 	copy(registeredServers, serversInfo.registeredServers)
 	serversInfo.RUnlock()
-	liveServers := 0
+
+	// Always have proxy.certRefreshConcurrency servers being refreshed in parallel
```
lifenjoiner: Setting to 1 equals to 1 by 1.

dnscrypt-proxy/serversInfo.go
```diff
@@ -227,13 +227,26 @@ func (serversInfo *ServersInfo) refresh(proxy *Proxy) (int, error) {
...
+	countChannel := make(chan struct{}, proxy.certRefreshConcurrency)
+	waitChannel := make(chan struct{})
 	var err error
-	for _, registeredServer := range registeredServers {
-		if err = serversInfo.refreshServer(proxy, registeredServer.name, registeredServer.stamp); err == nil {
-			liveServers++
-			proxy.xTransport.internalResolverReady = true
-		}
+	for i := range registeredServers {
+		// Would block if countChannel is already filled
```
lifenjoiner: "already filled" --> "full"?

keatonLiu: yes, I will correct that

dnscrypt-proxy/serversInfo.go
keatonLiu: Do you mean that setting to 1 is serial? How do I change the comment?

lifenjoiner: Yeah, one at a time. What about no comments?

keatonLiu: OK, that's fine, I will remove them

lifenjoiner:
LGTM.

lifenjoiner:
My previous negligence is fixed.

keatonLiu:
how about combine all the ifs together, which looks better

lifenjoiner:
> how about combine all the ifs together, which looks better

Good catch. But I'm not sure about the first 2.

lifenjoiner:
One more thing, the change from `Lock()` to `RLock()` will allow serving client requests, on one hand increases the availability, on the other hand could break the sorted latency list.

keatonLiu:
Do you mean that the list will be changed during output? I think RLock only allows concurrent reading, so I don't think the list will be changed during serving client requests. What's more, the sorted list is only about the initialRtt, which is only changed by calling the refresh function.

lifenjoiner:
Amend:
... could break the sorted latency log list, if the query log is also written to the same file (`stdout` for example).

Not saying the `serversInfo.inner`.

keatonLiu:
> Amend: ... could break the sorted latency log list, if the query log is also written to the same file (`stdout` for example).
> 
> Not saying the `serversInfo.inner`.

Oh, that's very detail, I haven't thought that. Do I need to change back?

lifenjoiner:
I'm not saying it's a bad thing. Let's see what @jedisct1 will say.

lifenjoiner:
@keatonLiu 
I reverted a commit to make this a atomic PR: One PR for one thing.
Those behavior changes are less related to this topic. They may worth a new PR.

lifenjoiner:
Thank you! You can submit other improvements within new PRs.

PR #2545, opened 2023-12-10
add timeout for udp and tcp dialer

keatonLiu:
The timeout can be set in the configuration file for the i/o timeout situation, so that sending and receiving DNS will end within a limited time. 
However, recently I found that  **_sometimes DNS resolution will block for significantly longer than the timeout_**, all block for **_more than two minutes_**.
The details is, when requesting certain relays which are not responding to the connection issued from the proxy, **_connection timeout will appear instead of i/o timeout_.** The timeout for connecting tcp is not set in the code. I haven't tried udp, I think they should all be configurable with timeouts.


jedisct1:
Thank you!

PR #2553, opened 2023-12-16
Listen `0.0.0.0` only on IPv4

lifenjoiner:
For https://github.com/DNSCrypt/dnscrypt-proxy/blob/f5912d7ca99cd30213e89e865bee5da3e8c88266/dnscrypt-proxy/example-dnscrypt-proxy.toml#L39-L40

Maybe golang's behavior on this had changed some time ago.
Currently, listen to `0.0.0.0` will be on both IPv4 and IPv6 interfaces, the same as `::`.
https://github.com/golang/go/blob/9b4b3e5acca2dabe107fa2c3ed963097d78a4562/src/net/ipsock_posix.go#L129

### DIFF
--- a/dnscrypt-proxy/coldstart.go
+++ b/dnscrypt-proxy/coldstart.go
@@ -122,11 +122,15 @@ func addColdStartListener(
 	listenAddrStr string,
 	captivePortalHandler *CaptivePortalHandler,
 ) error {
-	listenUDPAddr, err := net.ResolveUDPAddr("udp", listenAddrStr)
+	network := "udp"
+	if isDigit(listenAddrStr[0]) {
+		network = "udp4"
+	}
+	listenUDPAddr, err := net.ResolveUDPAddr(network, listenAddrStr)
 	if err != nil {
 		return err
 	}
-	clientPc, err := net.ListenUDP("udp", listenUDPAddr)
+	clientPc, err := net.ListenUDP(network, listenUDPAddr)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


lifenjoiner:
Or change the comments instead?

jedisct1:
Is this not relevant any more?

`0.0.0.0` should listen only to IPv4, this is the original behavior and what most people would expect.

lifenjoiner:
Didn't get any response, I guess you are expecting some better solution.
Frankly, my patch is not elegant, just simple and works, together with the context.
I don't get a more concise one. A wrapper should be common enough. That will be complicated and indirect.
If you don't opposite. I will solve the conflict and merge.

dnscrypt-proxy/coldstart.go
```diff
@@ -122,11 +122,15 @@ func addColdStartListener(
 	listenAddrStr string,
 	captivePortalHandler *CaptivePortalHandler,
 ) error {
-	listenUDPAddr, err := net.ResolveUDPAddr("udp", listenAddrStr)
+	network := "udp"
+	if isDigit(listenAddrStr[0]) {
```
jedisct1, line 126: This is a bit confusing. Maybe adding an `isIPv4()` function or using `! strings.HasPrefix(listenAddrStr, "[")` would be more self-explanatory.

lifenjoiner: 😂
explanatory Vs efficient

lifenjoiner: Appended 8d43ce9 introducing `isIPv4` as a variable.

PR #2605, opened 2024-03-26
Forwarding plugin: Support forwarding subdomains of the root domain

edmonds:
This commit updates the forwarding plugin to support matching subdomains of the root domain ("."). It looks like the forwarding plugin already performs subdomain matches against the domains specified in the forwarding rules files, but matches against the root domain weren't working because of the way matches are performed by comparing the normalized presentation format QNAME (which omits the trailing dot for all QNAMEs except the root domain name).

Without this commit, only queries where the QNAME is exactly "." would match a forwarding rule for the "." domain, like this (with `offline_mode = true` and a single forwarding rule for the "." domain):

```
[2024-03-25 21:13:31]	100.100.100.100	.	NS	FORWARD	0ms	127.0.0.1:53
[2024-03-25 21:13:36]	100.100.100.100	com	NS	NOT_READY	0ms	-
```

With this commit I get the expected result:

```
[2024-03-25 21:40:07]	100.100.100.100	.	NS	FORWARD	0ms	127.0.0.1:53
[2024-03-25 21:40:09]	100.100.100.100	com	NS	FORWARD	0ms	127.0.0.1:53
```

### DIFF
--- a/dnscrypt-proxy/plugin_forward.go
+++ b/dnscrypt-proxy/plugin_forward.go
@@ -89,8 +89,9 @@ func (plugin *PluginForward) Eval(pluginsState *PluginsState, msg *dns.Msg) erro
 		if candidateLen > qNameLen {
 			continue
 		}
-		if qName[qNameLen-candidateLen:] == candidate.domain &&
-			(candidateLen == qNameLen || (qName[qNameLen-candidateLen-1] == '.')) {
+		if (qName[qNameLen-candidateLen:] == candidate.domain &&
+			(candidateLen == qNameLen || (qName[qNameLen-candidateLen-1] == '.'))) ||
+			(candidate.domain == ".") {
 			servers = candidate.servers
 			break
 		}


jedisct1:
Thank you!

Just curious: what are use cases for forwarding everything? That seems to completely defeat the entire point of the software.

edmonds:
I want to use dnscrypt-proxy's filtering capabilities to provide a filtered DNS service to my LAN, but I already have a local full recursive DNS setup provided by Unbound. So I simply want to forward DNS queries from dnscrypt-proxy to Unbound without caching in dnscrypt-proxy. This also lets me have a unified cache for both the LAN clients that are querying Unbound directly vs. the clients that are querying the filtered dnscrypt-proxy service.

Yes it does avoid using most of the functionality offered by dnscrypt-proxy but it was very easy to set up :-)

edmonds:
I see the CI build failed but I'm not sure if it's related to the changes in this PR or not.

```
Warning: 26 18:53:07] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
Warning: 26 18:53:10] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
Warning: 26 18:53:13] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
Error: 3-26 18:53:13] [ERROR] No valid network configuration for [odohtarget]
Warning: 26 18:53:25] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
Warning: 26 18:53:25] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
*** Test #61 FAILED ***
Warning: 26 18:53:29] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
*** Test #62 FAILED ***
Warning: 26 18:53:41] [WARNING] Unable to decrypt response from [odohtarget]: [Malformed response]
```

PR #2625, opened 2024-05-03
Added DNS Stamp support for plain DNS over 53 (udp/tcp)

jdamick:
Refectored the _dnsExchange to create reusable udpExchange & tcpExchange 
Also added support for a "." forward rule (if this is acceptable)
Added associated unit tests

### DIFF
--- a/dnscrypt-proxy/dnsutils.go
+++ b/dnscrypt-proxy/dnsutils.go
@@ -434,79 +434,107 @@ func _dnsExchange(
 		if err != nil {
 			return DNSExchangeResponse{err: err}
 		}
-		udpAddr, err := net.ResolveUDPAddr("udp", serverAddress)
+		packet, rtt, err = udpExchange(proxy, serverAddress, relay, binQuery)
 		if err != nil {
 			return DNSExchangeResponse{err: err}
 		}
-		upstreamAddr := udpAddr
-		if relay != nil {
-			proxy.prepareForRelay(udpAddr.IP, udpAddr.Port, &binQuery)
-			upstreamAddr = relay.RelayUDPAddr
-		}
-		now := time.Now()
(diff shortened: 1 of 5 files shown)


jdamick:
I'm not sure that this failure is due to my changes?

jedisct1:
Errrr...

The promise of `dnscrypt-proxy` is that all the DNS traffic leaving the local network is authenticated and encrypted.

If suddenly, after an update, people silently get a mix of encrypted and unencrypted traffic, people will get really confused and lose trust in the software.

Sorry, but such a change cannot be merged. `dnscrypt-proxy`'s purpose is to support encrypted DNS traffic only.

jdamick:
Help me understand how it would silently change behavior? Wouldn't they have to create a plain dns stamp to support it?

jedisct1:
Servers are generally automatically and dynamically selected according to their latency. Unless they want anonymization, people don't have to care about what protocols are being used. Sources have the ability to publish stamps for unencrypted servers anytime they want, but users can be assured that `dnscrypt-proxy` is not going to use them.

"How do I know that I'm using encrypted transactions" is also a recurring question. Typical answers to it are:
- "stop the dnscrypt-proxy process. If you can't resolve anything any more, then you were using encrypted transactions because unencrypted DNS transactions never leave the local network when using dnscrypt-proxy"
- "browse that 3rd party website, and if the DNS resolvers are not your ISP's, then you are using DNS over secure channels"

ECH is also documented as safe to use with `dnscrypt-proxy` without having web browsers directly talk to DoH servers, because once again, outgoing connections are assumed to be always sent over secure channels.

Supporting plain DNS would invalidate all these claims.

`dnscrypt-proxy`'s purpose is to allow clients to connect to upstream servers using encrypted DNS protocols.

For other applications, traditional DNS caches such as `dnsmasq` are a better fit.

PR #2643, opened 2024-06-07
chore: fix some comments

cloudclaim:


jedisct1:
Thanks!

(but no need to patch what's in `vendor`, since this is maintained externally)

PR #2729, opened 2024-11-24
DNS-over-HTTP support, making TLS optional

lch361:
### Motivation

There are many situations where DNSCrypt-proxy can run under the reverse TLS proxy. In that case, built-in DNSCrypt-proxy TLS server becomes redundant.

### Solution

If the certificate and key are empty strings, instead of erroring out, start a server without TLS. Just pure HTTP.

### Example

You can configure nginx like this:
```nginx
http {
	ssl_certificate /etc/letsencrypt/live/whatever.net/fullchain.pem;
	ssl_certificate_key /etc/letsencrypt/live/whatever.net/privkey.pem;
	server {
		listen 443 ssl;
		root /srv/www;
		location /dns-query {
  			proxy_pass http://localhost:36153/;
		}
	}
}
```
And `dnscrypt-proxy.toml` like this:
```toml
[local_doh]
listen_addresses = ['[::1]:36153', '127.0.0.1:36153']
path = '/'
```
To get TLS from Nginx instead of dnscrypt-proxy.

### Trivia

Tested this patch on my server, everything works just as I intended. Hopefully this will make a fine addition to this project :)

### DIFF
--- a/dnscrypt-proxy/local-doh.go
+++ b/dnscrypt-proxy/local-doh.go
@@ -98,16 +98,20 @@ func (handler localDoHHandler) ServeHTTP(writer http.ResponseWriter, request *ht
 
 func (proxy *Proxy) localDoHListener(acceptPc *net.TCPListener) {
 	defer acceptPc.Close()
-	if len(proxy.localDoHCertFile) == 0 || len(proxy.localDoHCertKeyFile) == 0 {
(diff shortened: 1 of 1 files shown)


jedisct1:
DNSCrypt-proxy is a client. It’s not meant to be a server.

The local_doh feature is just a hack for a limitation in Firefox.

It’s not a replacement for doh-server.

PR #2783, opened 2025-02-10
Fix bootstrapResolvers IgnoreSystemDNS and typo

ACEx86:


ACEx86:
Also note that
queries will be sent to dnscrypt-proxy itself, if it is already
running with active servers (*)
or else, queries will be sent to fallback **(change toml to bootstrap?)** servers
finally, if `ignore_system_dns` is `false`, queries will be sent
to the system DNS

this doesn't apply to xtrasport because: 
```
if xTransport.ignoreSystemDNS {
	if xTransport.internalResolverReady {
		for _, proto := range protos {
			foundIP, ttl, err = xTransport.resolveUsingResolvers(proto, host, xTransport.internalResolvers)
			if err == nil {
				break
			}
		}
	} else {
		err = errors.New("Service is not usable yet")
		dlog.Notice(err)
	}
} else { <<--- Beggin
	foundIP, ttl, err = xTransport.resolveUsingSystem(host)
	if err != nil {
		err = errors.New("System DNS is not usable yet")
		dlog.Notice(err)
	}
}
if err != nil {
	for _, proto := range protos {
		if err != nil {
			dlog.Noticef(
				"Resolving server host [%s] using bootstrap resolvers over %s",
				host,
				proto,
			)
		}
		foundIP, ttl, err = xTransport.resolveUsingResolvers(proto, host, xTransport.bootstrapResolvers)
		if err == nil {
			break
		}
	}
}
if err != nil && xTransport.ignoreSystemDNS { <<----
	dlog.Noticef("Bootstrap resolvers didn't respond - Trying with the system resolver as a last resort")
	foundIP, ttl, err = xTransport.resolveUsingSystem(host)
}
```
Also if im not wrong if we are not ignoring system dns in xtransport you go by the wrong order.
System Dns is not ignored->System Dns. Instead of Resolver->Fail->Bootstrap->Fail->System Dns

I have to also add that maybe you haven't noticed this because when bootstrap is empty ignore system dns is always true and  if system dns fail at the beggin then you fall wrongly to ignore system dns if bootstrap fail because it is true.

lifenjoiner:
The workflow is https://github.com/DNSCrypt/dnscrypt-proxy/discussions/1851#discussioncomment-5568852.
Yes, this is only for downloading source files and then initializing DoH which without IP shipped.
It takes the System DNS as a last resort, if `ignoreSystemDNS` when all previous ways failed. Otherwise, the program may not start up.
That is bootstrap.

User's queries always use the encrypted ways as promised and documented.

ACEx86:
The system dns will fail anyway if the dnscrypt listen address is the only address configured. So in that case it is trying resolver if false but is allways true and then trying system dns because it is always true but that is wrong.
The default value of ignore system dns is true on xtransport and never set when you have empty bootstrap resolvers.
By the way as i said when you have empty bootstrap the default value of bootstrap is overwrited.
So resolver fails not ready, bootstrap never used, resolver fails again because it is the only system dns address configured because no md file.

dnscrypt-proxy/xtransport.go
```diff
@@ -447,7 +447,7 @@ func (xTransport *XTransport) resolveAndUpdateCache(host string) error {
 			}
 		}
 	}
-	if err != nil && xTransport.ignoreSystemDNS {
+	if err != nil && !xTransport.ignoreSystemDNS {
```
jedisct1, line 450: If `ignoreSystemDNS` is `false`, the system DNS is already tried just above.

Here, this is only if it is `true`, but for some reason, the configured bootstrap resolvers didn't work. Hence the warning message because this is not supposed to happen, and maybe the configuration needs to be changed.

ACEx86:
As for
Yes, this is only for downloading source files you mean the .md i guess and then initializing DoH which without IP shipped.
That is not true the xtrasport find the ip addresses without connections, even without internet connection unless it can't.

dnscrypt-proxy/config.go
```diff
@@ -396,9 +396,9 @@ func ConfigLoad(proxy *Proxy, flags *ConfigFlags) error {
 				return fmt.Errorf("Bootstrap resolver [%v]: %v", resolver, err)
 			}
 		}
-		proxy.xTransport.ignoreSystemDNS = config.IgnoreSystemDNS
```
jedisct1, line 399: If there are no bootstrap resolvers (?), using the system resolvers feels reasonable.

ACEx86:
@jedisct1 I need comment on the cache and about internal resolver ready.
Because it also seem like using the internal resolver anyway later if no system dns present and by that here i mean system dns is dnscrypt listen address

jedisct1:
It should already work if your system DNS resolver is the address `dnscrypt-proxy` listens on, since this is how most people are using it.

PR #2794, opened 2025-02-24
Fix invalid documentation with DNSSEC in the example config file.

virtualfuzz:
The comment says that server must support DNSSEC, but it sets require_dnssec = false 

I think require_dnssec needs to be set to true to require servers to support DNSSEC, and not the other way around.

### DIFF
--- a/dnscrypt-proxy/example-dnscrypt-proxy.toml
+++ b/dnscrypt-proxy/example-dnscrypt-proxy.toml
@@ -76,7 +76,7 @@ odoh_servers = false
 ## Require servers defined by remote sources to satisfy specific properties
 
 # Server must support DNS security extensions (DNSSEC)
-require_dnssec = false
+require_dnssec = true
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Servers must support DNSSEC when that filter is set to `true`, but by default, it is `false`.

The comment refers to what setting that filter to `true` does, not to what the default does.

But if this is confusing, we can add something like "if set to `true`" to the comment, and the other ones as well.

jedisct1:
There are booleans all over the place, including many others where the default value is `false`.

Adding `if set to true` everywhere feels counterproductive and doesn't improve clarity.

From all the examples, if feels pretty obvious, that `true` enables a property, and `false` disables it.
Server filters such as the one for DNSSEC are also [documented](https://github.com/DNSCrypt/dnscrypt-proxy/wiki/Configuration-Sources).

Status quo looks fine.

PR #2814, opened 2025-03-28
Reduce DHCP DNS detector instances

lifenjoiner:
Fix https://github.com/DNSCrypt/dnscrypt-proxy/discussions/2812.

Just starting the daemons is enough, as there will be messages indicating errors. Besides, some routers are not stable to response these requests.
https://github.com/DNSCrypt/dnscrypt-proxy/blob/67c121317701b01409279a312d4d06fd51123557/dnscrypt-proxy/plugin_forward.go#L193-L209


jedisct1:
Thank you!

PR #2843, opened 2025-05-08
Do not call deprecated rand.Seed

crrodriguez:
It is a noop in go >= 1.24 so this code no longer does what was intended.

### DIFF
--- a/dnscrypt-proxy/main.go
+++ b/dnscrypt-proxy/main.go
@@ -1,11 +1,8 @@
 package main
 
 import (
-	crypto_rand "crypto/rand"
-	"encoding/binary"
 	"flag"
 	"fmt"
-	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


jedisct1:
A no-op doesn't hurt either. And it can still be useful if people use old versions of go because this is what their Linux distribution ships with.

crrodriguez:
> A no-op doesn't hurt either. And it can still be useful if people use old versions of go because this is what their Linux distribution ships with.

currently this package requires go 1.24.. it won't build with older distributions unless you ignore go.mod requirements.

jedisct1:
And after that change you would need to change the go.mod requirements *and* revert that change. 

Removing the seeding doesn't improve anything, it can only reduce security. So, until the function actually disappears from the standard library, I'd rather keep the seeding.

PR #2846, opened 2025-05-18
Force the use of HTTPv3 with supplied hosts

junzhli:
Hello team!

I have been struggling with some hosts with httpv3 for a long while and I created feature that it will force the use of http v3 protocol for those hosts in a configured host list under toml file, which won't check in the existence of http3 support via the `alt-svc` header that contains `h3: <target port>` and presumably supports http3 protocol

It works by the supplied list of hosts under the `dnscrypt-proxy.toml`
```
## Force the use of http3 on the supplied hosts
force_http3_hosts = ['<host>[:<port>]'] # e.g. force_http3_hosts = ['dns.cloudflare.com']
```

I hope this to be included in future version for people who will also need this as some DoH servers actually supports v3 but it can't be used such as NextDNS

Thanks!

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -39,6 +39,7 @@ type Config struct {
 	UserName                 string         `toml:"user_name"`
 	ForceTCP                 bool           `toml:"force_tcp"`
 	HTTP3                    bool           `toml:"http3"`
+	ForceHTTP3Hosts          []string       `toml:"force_http3_hosts"`
 	Timeout                  int            `toml:"timeout"`
 	KeepAlive                int            `toml:"keepalive"`
 	Proxy                    string         `toml:"proxy"`
@@ -120,6 +121,7 @@ func newConfig() Config {
 		CertRefreshConcurrency:   10,
 		CertRefreshDelay:         240,
 		HTTP3:                    false,
+		ForceHTTP3Hosts:          []string{},
 		CertIgnoreTimestamp:      false,
 		EphemeralKeys:            false,
 		Cache:                    true,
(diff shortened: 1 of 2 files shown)


jedisct1:
This isn’t a very elegant workaround, and it wouldn’t be easy to use.

If the issue is that some providers don’t send the correct HTTP headers to indicate they support HTTP/3, maybe you could reach out to them—especially if it’s a company with customer support—and ask them to add these headers.

Alternatively, perhaps there could be an option to try HTTP/3 first, and fall back to HTTP/2 on error or timeout. That would be a global option and would introduce some startup delay, but at least it wouldn’t require many configuration changes, regardless of the network.

junzhli:
I understand it's not quite convenient at this moment. The reason that I propose this approach to explicitly use httpv3 by default is that I see Adguard supports the force use of httpv3 in similar way by introduced scheme name with uri `h3` (e.g. `h3://dns.cloudflare.com/dns-query`). Perhaps the `alt-svc` indicators on some popular ones are still missing, that's why it needs to be introduced as part of feature.

Btw, I agree with your proposal. It's better to have a global option that probes the http3 capability on target resolver in that way at startup! Maybe I can have a pull request in that way when I have time lol

jedisct1:
Inventing different schemes for the various HTTPS variants isn’t a good idea. I’ve never seen any software do that. It should simply be `https`. What will happen when servers start using HTTP/4?

And even without looking to the future, it’s common for UDP to be blocked, so depending on the network, HTTP/3 may not work at all. That’s not something we have to worry about with regular HTTP clients, so it shouldn’t be any different here.

junzhli:
I understand your concern and idea. I just created another pr which eases the configuration part and is replaced with your idea like probing http3 capability at startup
https://github.com/DNSCrypt/dnscrypt-proxy/pull/2848

junzhli:
> Inventing different schemes for the various HTTPS variants isn’t a good idea. I’ve never seen any software do that. It should simply be `https`. What will happen when servers start using HTTP/4?
> 
> And even without looking to the future, it’s common for UDP to be blocked, so depending on the network, HTTP/3 may not work at all. That’s not something we have to worry about with regular HTTP clients, so it shouldn’t be any different here.

Just fyi, I also saw some similar mechanism in `curl` tool, which provides some args `--http3-direct` that allow to directly attempt to establish http3 connection rather than using `alt-svc` approach over http2 in the first attempt. it might still worth adopting the current model
https://daniel.haxx.se/blog/2019/08/05/first-http-3-with-curl/

PR #2866, opened 2025-05-24
make top domains sorting stable

ghost:
Make at least top domains sorting stable as it is the one which most likely contains duplicate counts that cause unnecessary and annoying movements in the dashboard.


jedisct1:
Thanks!

PR #2871, opened 2025-05-31
Monitoring UI: fix UI scrolling issues

lifenjoiner:


lifenjoiner:
A strange thing observed:
On 00914a4e, the offset scrolled to is not correct, if the `api/ws` connection is recreated (a new query is finished, and the metric changes), on both Windows Firefox and Edge. Patches 91e20e4, #2871, and #2872 can "fix" it.

What does it happen? Maybe there is something interesting.

lifenjoiner:
Maybe it is because of:
When `api/ws` connection is recreated, there is also a `api/metrics` request. Both get data, and trigger the page to redo rendering. But, the interval is dramatically short, neither too short nor too long, the page scrolls while a new rendering is happening.

lifenjoiner:
Finally:
There is a pong message received for each new `api/ws` connection, but without metrics data.

PR #2872, opened 2025-05-31
Web UI: fix duplicate metric updates after WebSocket reconnection

ghost:
Stops the polling script after a successful WebSocket reconnection.


ghost:
I guess you want to give `newWs` a better name and re-add a comment like `// Reset global variable ws`?

PR #2873, opened 2025-05-31
Always redo SetWriteDeadline for each broadcast

lifenjoiner:
Fix timeout on api/ws


lifenjoiner:
`SetWriteDeadline` has been called at the beginning. We need to reset the deadline.

PR #2878, opened 2025-06-05
Empty signal handler for Windows

lifenjoiner:
Save some resources.
CPU, memory: not measured
exe size: -4K


lifenjoiner:
There is no SIGHUP on Windows.

PR #2888, opened 2025-06-08
improve (stabilize) server rtt computation

ghost:
prevent sporadic outliers from heavily affecting the server rtt

e.g. the new scoring system stops considering the server rtt when it's over 1s: just because a fast server has 2 slow responses in a row shouldn't put it in the bad server category

on the other hand this change still allows to adapt to actual slowdowns quite quickly

### DIFF
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	RTTEwmaDecay = 10.0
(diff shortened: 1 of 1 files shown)


jedisct1:
Instead of magic constants, maybe define something like

```go
 const (
      RTTEwmaDecay         = 20.0
      RTTOutlierUpperBound = 10.0
      RTTOutlierLowerBound = 0.1
  )
```

Also, indentation is funky.

PR #2898, opened 2025-06-25
sample config: update quad9 source URLs

c4rlo:
The existing quad9-resolvers.md URL no longer works. I found https://github.com/Quad9DNS/dnscrypt-settings with more up-to-date info.


jedisct1:
https://quad9.net/dnscrypt/quad9-resolvers.md still seems to work for me.

c4rlo:
That's the new URL. https://www.quad9.net/quad9-resolvers.md is the old URL that no longer works.

PR #2930, opened 2025-08-15
Rename PatternMatcher's field names and improve pattern rule examples

lifenjoiner:
* "block" initiated and extended to allowed and cloaking names
* `\-` stands for '-' in bracket expressions, requires `\`
* `_` is not allowed in host names, but allowed in DNS for special purpose


jedisct1:
Makes sense. Thank you!

PR #2948, opened 2025-09-07
Skip getting a server candidate if unnecessary

lifenjoiner:
Reduce locks.

Passing a `*Proxy` seems not elegant. But, it works, and avoids redoing unpack and repack the packets to add EDNS0 Padding outside `ApplyQueryPlugins`.


jedisct1:
Avoiding passing `*Proxy` around would be better

An alternative could be to pass a `getServerInfo` function pointer to the plugins, so that the function is called only when necessary.

PR #2950, opened 2025-09-07
Reduce lock acquisition

lifenjoiner:
Unlock ASAP.
`responsePlugins` can be empty.


lifenjoiner:
Nit picking, maybe.

I'm diagnosing and identifying a potential issue as the picture shows, which happens at the restore of PC hibernation, high volume client queries while server refreshing.
longer delay

PR #2954, opened 2025-09-08
Enable Firefox plugin only when local DOH is enabled

lifenjoiner:


jedisct1:
I don't use Firefox and I don't live in the US where it silently forces people to use Cloudflare, but my understanding is that the Firefox plugin is always required. Without this, Firefox will automatically use its own DoH client, especially if no DoH servers have been configured.

welwood08:
Correct, the purpose of the Firefox plugin is to present a canary domain for Firefox to detect that signals it to disable DoH. This plugin is not technically necessary when Firefox is using local DoH (because the user is expected to manually configure Firefox's DoH settings to point to `dnscrypt-proxy`) but very necessary when not using local DoH (because Firefox may automatically bypass `dnscrypt-proxy` using external DoH servers). Perhaps there needs to be more description/comments to explain this in the code in case that link [in the comment at the top of the Firefox plugin](https://github.com/DNSCrypt/dnscrypt-proxy/blob/c474965454543f6a47f05684b33061cfed6f219d/dnscrypt-proxy/plugin_firefox.go#L1) breaks.

lifenjoiner:
Oh~ OK! I don't use DOH in Firefox either. This is really weird. Forcibly interrupting Firefox using DOH is a bit weird too.

I was wondering if user controlled rules work and just tested:
1. Use blocking rules. It should work, if users don't set `blocked_query_response` to IPs.
2. Use cloaking rules like `use-application-dns.net doesnotexist.com`, but failed. It requires there must be no error. The rule is ignored. Is it right?

BTW:
https://github.com/DNSCrypt/dnscrypt-proxy/blob/064d7ac7b3657a4a51026a257841c142baca92a7/dnscrypt-proxy/plugin_cloak.go#L258
This can lead user queries to escape the encryption channels and can be improved, especially on a router. If anyone doesn't dive in, I'll have a try.

lifenjoiner:
Append:
> Note: The canary domain only applies to users who have DoH enabled as the default option. It does not apply for users who have made the choice to turn on DoH by themselves.

https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet

PR #2957, opened 2025-09-09
Enable `systemctl reload` to send reload signal

welwood08:
Should resolve #2955

I'm not in a position to properly test my changes still but this looks simple enough with no obvious side-effects. I followed the upstream service module's [logging example](https://github.com/kardianos/service/blob/b7ded6fcb7497ba4f88edf9cc3445411175b8897/example/logging/main.go#L68) closely. A quick search of the upstream code suggests this option only has any effect in the context of SystemD on Linux and [will add an appropriate `ExecReload=` line](https://github.com/kardianos/service/blob/b7ded6fcb7497ba4f88edf9cc3445411175b8897/service_systemd_linux.go#L317) to the generated service file.

### DIFF
--- a/dnscrypt-proxy/main.go
+++ b/dnscrypt-proxy/main.go
@@ -77,12 +77,15 @@ func main() {
 		flags: &flags,
 	}
 
+	svcOptions := make(service.KeyValue)
+	svcOptions["ReloadSignal"] = "HUP"
 	svcConfig := &service.Config{
 		Name:             "dnscrypt-proxy",
 		DisplayName:      "DNSCrypt client proxy",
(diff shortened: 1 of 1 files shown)


Winterhuman:
EDIT: Nevermind, `svcOptions["ReloadSignal"]` is different to systemd's `ReloadSignal=`; kardianos's service would add a `ExecReload=` line as expected

jedisct1:
@Winterhuman So, it that change correct? Does it work for you ?

Winterhuman:
Oh right, yeah, the change is correct. Sorry for not being clear

jedisct1:
Thanks!

PR #2959, opened 2025-09-09
Improve target CNAME resolving for cloaking

lifenjoiner:
Avoid bypass and increase privacy.

Issues with CNAME cloaking rules:
* Rule bypass if upstream server fails
* Privacy leak if `net.LookupIP`/`net.LookupHost` flows to plain DNS
* Out of service if clients query AAAA records but IPv6 connection is not available

### DIFF
--- a/dnscrypt-proxy/systemd_linux.go
+++ b/dnscrypt-proxy/systemd_linux.go
@@ -9,6 +9,15 @@ import (
 	"github.com/jedisct1/dlog"
 )
 
+func hasStringElement(array []string, s string) bool {
+	for _, ele := range array {
+		if ele == s {
+			return true
+		}
+	}
+	return false
+}
+
 func (proxy *Proxy) addSystemDListeners() error {
 	files := activation.Files(true)
 
@@ -19,15 +28,22 @@ func (proxy *Proxy) addSystemDListeners() error {
 			)
 		}
 		dlog.Warn("Systemd sockets are untested and unsupported - use at your own risk")
+		proxy.listenAddresses = make([]string, 0)
 	}
 	for i, file := range files {
 		defer file.Close()
+		var listenAddress string
 		if listener, err := net.FileListener(file); err == nil {
 			proxy.registerTCPListener(listener.(*net.TCPListener))
-			dlog.Noticef("Wiring systemd TCP socket #%d, %s, %s", i, file.Name(), listener.Addr())
+			listenAddress = listener.Addr().String()
+			dlog.Noticef("Wiring systemd TCP socket #%d, %s, %s", i, file.Name(), listenAddress)
 		} else if pc, err := net.FilePacketConn(file); err == nil {
 			proxy.registerUDPListener(pc.(*net.UDPConn))
-			dlog.Noticef("Wiring systemd UDP socket #%d, %s, %s", i, file.Name(), pc.LocalAddr())
+			listenAddress = pc.LocalAddr().String()
+			dlog.Noticef("Wiring systemd UDP socket #%d, %s, %s", i, file.Name(), listenAddress)
+		}
+		if len(listenAddress) > 0 && !hasStringElement(proxy.listenAddresses, listenAddress) {
+			proxy.listenAddresses = append(proxy.listenAddresses, listenAddress)
 		}
 	}
 	return nil
--- a/dnscrypt-proxy/systemd_linux.go
+++ b/dnscrypt-proxy/systemd_linux.go
@@ -4,19 +4,11 @@ package main
 
 import (
 	"net"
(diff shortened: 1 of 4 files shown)


jedisct1:
What does that do exactly?

lifenjoiner:
* Avoid cloaking rule bypass
https://github.com/DNSCrypt/dnscrypt-proxy/blob/e9d11c5d54c546f61a281ca83c271ae862b7a637/dnscrypt-proxy/example-cloaking-rules.txt#L12
If there is an unexpected error from `net.LookupIP`, the rule will be bypassed, silently.

* Improve target CNAME resolving for cloaking
https://github.com/DNSCrypt/dnscrypt-proxy/blob/e9d11c5d54c546f61a281ca83c271ae862b7a637/dnscrypt-proxy/example-dnscrypt-proxy.toml#L444
When doing CNAME flattening. `net.LookupIP` uses system DNS. If dnscrypt-proxy is serving on a server (router?) which does not use dnscrypt-proxy as the system DNS, doesn't it use the plain DNS?

jedisct1:
To be honest, I'm super scared about touching anything in `xtransport`, as any functional changes there can have too many implications.

lifenjoiner:
We test it and improve it.

This only reuses the host name resolving part for DoH stamps.

It works for both:
```
[s] [NOTICE] dnscrypt-proxy service is not usable yet
[s] [NOTICE] Resolving server host [dns.nextdns.io] using bootstrap resolvers over udp
[s] [NOTICE] [dnscry.pt-lasvegas-ipv4] OK (DNSCrypt) - rtt: 138ms
[s] [NOTICE] [dnscry.pt-tokyo-ipv4] OK (DNSCrypt) - rtt: 232ms
[s] [INFO] [nextdns-ultralow] TLS version: 304 - Protocol: h2 - Cipher suite: 4865
[s] [NOTICE] [nextdns-ultralow] OK (DoH) - rtt: 262ms
[s] [NOTICE] Sorted latencies:
[s] [NOTICE] -   138ms dnscry.pt-lasvegas-ipv4
[s] [NOTICE] -   232ms dnscry.pt-tokyo-ipv4
[s] [NOTICE] -   262ms nextdns-ultralow
[s] [NOTICE] Server with the lowest initial latency: dnscry.pt-lasvegas-ipv4 (rtt: 138ms)
[s] [NOTICE] dnscrypt-proxy is ready - live servers: 3
[s]   127.0.0.1       forcesafesearch.google.com      A       PASS    157ms   dnscry.pt-lasvegas-ipv4
[s]   127.0.0.1       www.google.com  A       CLOAK   161ms   -
[s] [NOTICE] Stopped.
```

jedisct1:
There are too many changes to xTransport here.

lifenjoiner:
Made some mistakes. I'm amending it.

lifenjoiner:
I added descriptions to commits. And here is a summary of key points:
* CNAME cloaking rule bypass: parental control bypass
* `net.LookupIP`/`net.LookupHost` may use plain DNS: privacy leak
* `net.LookupIP`/`net.LookupHost` only queries IPv4 via IPv4 and IPv6 via IPv6: won't get IPv6 if ...
* Scenarios: Cross-device services
* Query and cache IPv4 and IPv6 separately
* Read locking covers reading the cache
* Upstream multi-tier CNAME resolving needs more time
* Update `internalResolvers` for systemd socket mode
* Align `resolve*` function interfaces, and adapt them for reuse
* Compact `resolveUsingResolver`

lifenjoiner:
Issues with CNAME cloaking rules:
* Rule bypass if upstream server fails
* Privacy leak if `net.LookupIP`/`net.LookupHost` flows to plain DNS
* Out of service if clients query AAAA records but IPv6 connection is not available

Reasoning of the patches:
Privacy ==> Use `internalResolvers` ==> Reuse `resolveUsingResolvers()` ==> Return `[]IP` ==> Adapt and unify `resolve*` functions.
Logic and results for `resolveAndUpdateCache` keep working.

Does it have any regressions?

dnscrypt-proxy/systemd_linux.go
```diff
@@ -9,6 +9,15 @@ import (
...
 
+func hasStringElement(array []string, s string) bool {
+	for _, ele := range array {
+		if ele == s {
+			return true
+		}
+	}
+	return false
+}
+
```
jedisct1: You can use `slices.Contains` for this.

lifenjoiner: Thank you!
I thought using a self written function in this less tasks case is better than importing a new package.

jedisct1:
Thanks!

PR #2963, opened 2025-09-14
Increase log timeout limit

lifenjoiner:
Indicates a little more information.


62f1346c-4576-42b3-afcb-23ef1820da34:
Maybe change this in the Monitoring UI too, for consistency?

PR #2971, opened 2025-09-22
Do not truncate buffer size of client UDP datagrams

MegaManSec:
Previously, handleColdStartClient would error out due to a truncated legitimate maximum-size EDNS0 query.


jedisct1:
Thanks! Yes, I don't remember why the `-1` was here.

LLMs are great :)

MegaManSec:
>LLMs are great:)

they can be! and then when reviewing code/triaging on an airplane with no internet, things can go not so great;)

jedisct1:
qwen3-coder-30b-a3b  doesn't work too bad locally. But yeah, there's still a lot manual triaging/review to be done afterwards.

PR #2972, opened 2025-09-22
Fix inverted nil-check in cache file resolution

MegaManSec:


dnscrypt-proxy/sources.go
```diff
@@ -124,7 +124,7 @@ func (source *Source) updateCache(bin, sig []byte) {
 	now := getCurrentTime()
 	file := source.cacheFile
 	absPath := file
-	if resolved, err := filepath.Abs(file); err != nil {
+	if resolved, err := filepath.Abs(file); err = nil {
```
jedisct1: `err = nil` ?

That shouldn't even compile.

MegaManSec: typo. fixed

jedisct1:
Thanks!

PR #2973, opened 2025-09-22
Do not leak a a UDP socket if quic.DialEarly fails

MegaManSec:


jedisct1:
That would cause a runtime panic, since `conn` would be `nil`.

PR #2974, opened 2025-09-22
Do not double-append system resolver list when using IPV6.

MegaManSec:
To16() returns a non-nil 16-byte form for both IPv4 and IPv6 addresses,so IPv4 addresses also satisfy the condition `ipv6 := foundIP.To16(); ipv6 != nil`. As a result, if useIPv6 is true and useIPv4 is false IPv4 addresses will still be appended to ips and can be returned, and if both flags are true IPv4 addresses can be appended twice.  This logic violates the configured IPv4/IPv6 preferences and biases the selection toward IPv4.


jedisct1:
Thanks a lot for pointing out that `to16()` could return IPv5 mapped address!

That was more generally addressed in cb1cd83b094b85bf6e2400b58b663d5f7c2d9c31 and afab24a187d7ece5bb19c7eafa9a61b6f68a98a1

Good catch!

PR #2975, opened 2025-09-22
Fix index-out-of-range panic in DNS message with empty Question section

MegaManSec:


jedisct1:
How does that happen? In ApplyQueryPlugins, queries that don't have qlen!=1 are rejected.

MegaManSec:
Indeed; apologies, I missed that.

PR #2976, opened 2025-09-22
Ensure Bootstrap and Explicit resolvers is not empty in plugin_forward

MegaManSec:


jedisct1:
These cases cannot occur.

We ensure that `bootstrapResolvers` is not empty when `$BOOTSTRAP` is used.  And we also the servers array always contains at least one valid server.

PR #2977, opened 2025-09-22
Reset dnsId on retries

MegaManSec:


jedisct1:
I don't understand. Why is that necessary?

MegaManSec:
It's just a correctness nit. Each retry is a new DNS transaction, so the dnsid/tid should seemingly be unique per connection; avoiding situations where late or reordered or duplicated replies from a previous attempt can still be accepted even if the server has been changed (unlike,y I know, but still).

jedisct1:
Since the question doesn't change, receiving a late answer is fine. This is exactly what high performance DNS caches do in order to use a late answer coming before the one from a retry.  But in `dns.Exchange`, such optimization is not made anyway. So, why not, even though we're just sending queries to localhost.

PR #2978, opened 2025-09-22
Treat NXDOMAIN as a valid answer, and SERVFAIL as invalid, per RFC 2308

MegaManSec:
RFC 2308 states that NXDOMAIN answers should be handled as a authorative negative response, and therefore are cacheable. As such, there is no need to retry other servers when this response is seen.  SERVFAIL should be treated as an invalid answer, which is retried.


jedisct1:
Nope. This is intentional.

MegaManSec:
Why? Shouldn't NXDOMAIN should be negatively cached?

jedisct1:
The forward plugin was made to forward internal domains to the local router. That registers hosts and devices as they appear on the network. If a query is made for a device when the computer starts and it is cached, when the device is plugged in, it may not appear. That results is a bad experience.

MegaManSec:
Thanks for the explanation.

PR #2979, opened 2025-09-22
Set http2Transport.ReadIdleTimeout and AllowHTTP correctly

MegaManSec:
Avoids a potential panic, too.

### DIFF
--- a/dnscrypt-proxy/xtransport.go
+++ b/dnscrypt-proxy/xtransport.go
@@ -280,7 +280,7 @@ func (xTransport *XTransport) rebuildTransport() {
 		}
 	}
 	transport.TLSClientConfig = &tlsClientConfig
-	if http2Transport, err := http2.ConfigureTransports(transport); err != nil {
+	if http2Transport, err := http2.ConfigureTransports(transport); err == nil {
 		http2Transport.ReadIdleTimeout = timeout
 		http2Transport.AllowHTTP = false
 	}


dnscrypt-proxy/xtransport.go
```diff
@@ -280,7 +280,7 @@ func (xTransport *XTransport) rebuildTransport() {
 		}
 	}
 	transport.TLSClientConfig = &tlsClientConfig
-	if http2Transport, err := http2.ConfigureTransports(transport); err != nil {
+	if http2Transport, err := http2.ConfigureTransports(transport); err == nil {
```
jedisct1, line 283: We want to disable HTTP, and set the timeout if http2 works. So, if `err` is `nil`.

jedisct1:
My bad, your static analyzer is actually correct.

MegaManSec:
btw, I just noticed when looking at this patch the code right above: https://github.com/MegaManSec/dnscrypt-proxy/blob/36e342bb71ae28d94fb34b17660505e55af9df32/dnscrypt-proxy/xtransport.go#L276 -- is this doing what it's supposed to? Why wouldn't it be just `if overrideCipherSuite { ... }`?
afaict, current code basically always downgrades unless the user list is every tls version up to 1.2 that Go knows about, right?

jedisct1:
> btw, I just noticed when looking at this patch the code right above: https://github.com/MegaManSec/dnscrypt-proxy/blob/36e342bb71ae28d94fb34b17660505e55af9df32/dnscrypt-proxy/xtransport.go#L276 -- is this doing what it's supposed to? Why wouldn't it be just `if overrideCipherSuite { ... }`? afaict, current code basically always downgrades unless the user list is every tls version up to 1.2 that Go knows about, right?

This is right. Because the cipher suite is ignored when using TLS 1.3. So, if people want specific cipher suites that are not the default ones, only TLS 1.2 can do it.

PR #2986, opened 2025-09-24
Better handle server failures and stale responses

MegaManSec:
The previous error handling logic could fail to correctly log a server failure, particularly during DoH TLS handshake errors (on `tls == nil || !tls.HandshakeComplete`) or when serving a stale response. This led to inaccurate server reliability scores. This also cleans up `serverInfo.noticeFailure` calls, and a single missing `serverInfo.noticeBegin` call in `processODoHQuery`.

### DIFF
--- a/dnscrypt-proxy/query_processing.go
+++ b/dnscrypt-proxy/query_processing.go
@@ -88,7 +88,10 @@ func processDNSCryptQuery(
 	if err != nil {
 		if stale, ok := pluginsState.sessionData["stale"]; ok {
 			dlog.Debug("Serving stale response")
-			response, err = (stale.(*dns.Msg)).Pack()
+			serverInfo.noticeFailure(proxy)
+			if staleResponse, packErr := (stale.(*dns.Msg)).Pack(); packErr == nil {
+				return staleResponse, nil
+			}
 		}
 	}
 
@@ -122,9 +125,12 @@ func processDoHQuery(
 
 	// Check for stale response if there was an error
 	if err != nil || tls == nil || !tls.HandshakeComplete {
+		serverInfo.noticeFailure(proxy)
 		if stale, ok := pluginsState.sessionData["stale"]; ok {
-			dlog.Debug("Serving stale response")
-			return (stale.(*dns.Msg)).Pack()
+			dlog.Debug("Serving stale response after error")
+			if staleResponse, err := (stale.(*dns.Msg)).Pack(); err == nil {
+				return staleResponse, nil
+			}
 		}
 	}
 
@@ -132,7 +138,6 @@ func processDoHQuery(
 	if err != nil {
 		pluginsState.returnCode = PluginsReturnCodeNetworkError
 		pluginsState.ApplyLoggingPlugins(&proxy.pluginsGlobals)
-		serverInfo.noticeFailure(proxy)
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


MegaManSec:
By the way, this code still contains a small bug:

```
            exchangeResponse, err := handleDNSExchange(proxy, serverInfo, &pluginsState, query, serverProto)

            // Update server statistics for WP2 strategy
            success := (err == nil && exchangeResponse != nil)
            proxy.serversInfo.updateServerStats(serverName, success)

```

Upon using a stale response, `err == nil`, so the serverstats don't get updated.

Maybe adding a `PluginsReturnCodeStale` and returning that when a stale response is used, but still using the response, would be appropriate?

jedisct1:
I'm not sure that we should call `serverInfo.noticeFailure` when we serve a stale response.
The server should have a chance to quickly recover from temporary lag or packet loss.

dnscrypt-proxy/query_processing.go
```diff
@@ -119,29 +123,30 @@ func processDoHQuery(
...
+	serverInfo.noticeFailure(proxy)
 
-	// Restore the original transaction ID
-	response := serverResponse
-	if len(response) >= MinDNSPacketSize {
-		SetTransactionID(response, tid)
+	// Attempt to serve a stale response as a fallback.
+	if stale, ok := pluginsState.sessionData["stale"]; ok {
+		dlog.Debug("Serving stale response after error")
+		if staleResponse, packErr := (stale.(*dns.Msg)).Pack(); packErr == nil {
```
jedisct1, line 141: Why `Pack()` here? (unlike for DNSCrypt)

dnscrypt-proxy/query_processing.go
```diff
@@ -156,10 +161,13 @@ func processODoHQuery(
 		return nil, nil
 	}
 
+	serverInfo.noticeBegin(proxy)
+
 	target := serverInfo.odohTargetConfigs[rand.Intn(len(serverInfo.odohTargetConfigs))]
 	odohQuery, err := target.encryptQuery(query)
 	if err != nil {
 		dlog.Errorf("Failed to encrypt query for [%v]", serverInfo.Name)
+		serverInfo.noticeFailure(proxy)
```
jedisct1: That doesn't seem to be a server error, so the server shouldn't be penalized.

dnscrypt-proxy/query_processing.go
```diff
@@ -86,8 +87,11 @@ func processDNSCryptQuery(
 	// Check for stale response if there was an error
 	if err != nil {
 		if stale, ok := pluginsState.sessionData["stale"]; ok {
-			dlog.Debug("Serving stale response")
-			response, err = (stale.(*dns.Msg)).Pack()
+			dlog.Debug("Serving stale response after error")
```
jedisct1: Stale responses are always served after error by definition :)

MegaManSec: got it

dnscrypt-proxy/query_processing.go
MegaManSec: it was being returned here: https://github.com/DNSCrypt/dnscrypt-proxy/blob/55fc36dbd7b3bc65230f82bcf2954d879b698ab6/dnscrypt-proxy/query_processing.go#L126

PR #2987, opened 2025-09-24
Log on pluginsState.ApplyLoggingPlugins error

MegaManSec:


jedisct1:
Do you need that?

I never did, since plugins already log things that are worth logging themselves.

MegaManSec:
i more had in mind the case that

```
    if questionMsg == nil {
        return errors.New("Question not found")
    }
```

occurs (if it's even possible). Perhaps it would be cleaner to just add a single message there, instead?

jedisct1:
I rather think that we already have way too many debug statements.

They are code bloat, that users have to pay for, both in code size and wasted CPU cycles.

They are only useful for developers, so we should ultimately only keep the ones that we *actually* need to understand common issues.

MegaManSec:
understood; seems reasonable

PR #2989, opened 2025-09-24
snapshot ipv4/ipv6/PTR cloakedNames under RLock to avoid race condition

MegaManSec:


MegaManSec:
nope, this was fixed already, didn't notice: https://github.com/DNSCrypt/dnscrypt-proxy/blob/master/dnscrypt-proxy/plugin_cloak.go#L306

PR #2991, opened 2025-09-24
defer clientsCountDec() to prevent counter leak on potential panic

MegaManSec:
Shouldn't actually happen, but I think it's worth it just in case.

P.S: this is my last PR:)

### DIFF
--- a/dnscrypt-proxy/plugin_dns64.go
+++ b/dnscrypt-proxy/plugin_dns64.go
@@ -93,6 +93,7 @@ func (plugin *PluginDNS64) Eval(pluginsState *PluginsState, msg *dns.Msg) error 
 	if !plugin.proxy.clientsCountInc() {
 		return errors.New("Too many concurrent connections to handle DNS64 subqueries")
 	}
+    defer plugin.proxy.clientsCountDec()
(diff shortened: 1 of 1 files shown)


dnscrypt-proxy/plugin_dns64.go
```diff
@@ -93,6 +93,7 @@ func (plugin *PluginDNS64) Eval(pluginsState *PluginsState, msg *dns.Msg) error
 	if !plugin.proxy.clientsCountInc() {
```
lifenjoiner, line 93: Step into this. The counter will not increase, if reaches the up bound. AI without review?

MegaManSec: I don't see how the upper bound is directly relevant to the change here. It's about processIncomingQuery and how if a panic occurs somewhere in that, the counter is never from deincremented. If that happens over and over, the server will be inoperable after the never-deincremented variable reaches the limit of clientscount (yes, the upper bound)

lifenjoiner: If panic, won't the program crash?

Strange things happened in my mind about this `defer`. 😂

PR #2993, opened 2025-09-25
Call `updateMonitoringMetrics` in `ApplyLoggingPlugins`

lifenjoiner:
SERVER_TIMEOUT, FORWARD, DROP and etc. results, for example.


jedisct1:
Would it make sense to call updateMonitoringMetrics() directly in ApplyLoggingPlugins ?

PR #3002, opened 2025-10-07
add combined EDNS IDs plugin

illia-malachyn:
Introduce PluginDeviceAndSubscriberID that injects both EDNS0 local options into outgoing DNS queries:
- device_id (code 65001)
- subscriber_id (code 65075)

Behavior:
- Adds an OPT record if missing and appends options only when BOTH IDs are configured.
- Prevents duplicates if options already exist.

Config changes:
- Enforce strict validation for `edns_device_id` and `edns_subscriber_id` as UUIDv4 without dashes.
- Only configure IDs when both are non-empty and valid.
- Sample config updated to document the expected format.


jedisct1:
This is confusing. These identifiers belong to a reserved private range, they don’t seem to be documented anywhere, and none of the open-source server implementations I know support them.

Also, these changes haven’t been discussed anywhere before. Was this PR intended for a private fork?

jedisct1:
Arbitrary key/value pairs can already be sent using the `query_meta` mechanism.

And `encrypted-dns-server` supports this natively for access control.

I’d recommend using the existing mechanism.

illia-malachyn:
> Also, these changes haven’t been discussed anywhere before. Was this PR intended for a private fork?

I'm sorry. Indeed, these changes are required for a private fork

PR #3026, opened 2025-11-10
Show number of goroutines if reaches max clients

lifenjoiner:
Need it to reveal info about the workload.

For example:
```
[2025-11-10 08:07:40] [WARNING] Too many incoming connections (max=500), number of goroutines: 806
[2025-11-10 08:07:40] [WARNING] Too many incoming connections (max=500), number of goroutines: 807
[2025-11-10 08:07:40] [WARNING] Too many incoming connections (max=500), number of goroutines: 807
```


jedisct1:
That's an implementation detail, and users shouldn't have to learn what a goroutine is.

So, maybe that would be better in DEBUG level.

lifenjoiner:
OK. Seperated `Debugf` lines added.

PR #3027, opened 2025-11-10
Update vendor

syphyr:
After running `go mod vendor` and `go mod tidy`, these additional files seem to have been left out.


jedisct1:
Done, thanks!

PR #3034, opened 2025-11-14
ci-test.sh: show line number where failed

lifenjoiner:
Easier to locate the failed test.

### DIFF
--- a/.ci/ci-test.sh
+++ b/.ci/ci-test.sh
@@ -13,7 +13,7 @@ t() {
 }
 
 fail() (
-    echo "*** Test #${TEST_COUNT} FAILED ***" >&2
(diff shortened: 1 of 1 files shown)


jedisct1:
Never heard about `$caller`. It that specific to some shell?

welwood08:
I believe `caller` (not `$caller`) is a [Bash builtin command](https://ss64.com/bash/caller.html). Whilst it looks useful, I don't think it's appropriate to assume it exists in a POSIX sh script.

lifenjoiner:
I only have msys2. Shall we use `$LINENO` as an argument to `fail()`?

lifenjoiner:
Updated to use `$LINENO` for compatibility.

PR #3036, opened 2025-11-14
Optimize for wake-up from hibernation

lifenjoiner:
1. Give time to NIC to be ready.
2. Avoid peak request times.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -324,7 +324,7 @@ func (proxy *Proxy) StartProxy() {
 				if liveServers == 0 {
 					delay = proxy.certRefreshDelayAfterFailure
 				}
-				clocksmith.Sleep(delay)
+				sleepIncludeInertia(delay)
 				liveServers, _ = proxy.serversInfo.refresh(proxy)
 				if liveServers > 0 {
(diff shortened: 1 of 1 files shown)


jedisct1:
We can probably just use `clocksmith.SleepWithGranularity` with a granularity larger than the default (5 seconds) instead.

lifenjoiner:
... uh, the difference:
clocksmith: always try to wake up in time;
`sleepIncludeInertia`: always try to sleep an "extra" minor duration, like "sleep in".

For cases that have hibernated beyond the app scheduled time, `sleepIncludeInertia` is better?
Draw a diagram and you'll get it. I'm not sure if the following is clear enough.
```
clocksmith:
|-------*-------+------+--------------*------------+------|
                     ^ ^                                  ^
                     1 2                                  3

sleepIncludeInertia:
|-------*-------+------+--------------*------------+------|
                     ^ ^                           ^
                     1 2                           3

1: frozen
2: wake up from hibernation, jump to 3
3: now
```

lifenjoiner:
Supplement:
The difference lies in the probability distribution of the delay.
We expect it to be as close as possible to the specified value, which is neither too large nor too small.
Devices: old Vs new;
Hibernation: very long Vs normal;
Customized sources or less servers specified.

PR #3067, opened 2025-12-24
Fix potential nil and panics 

ACEx86:


welwood08:
From just a quick look, it seems most of these changes don't actually fix panics, just move them to later in the functions. How many of these were suggested by AI?

dnscrypt-proxy/xtransport.go
```diff
@@ -487,20 +487,22 @@ func (xTransport *XTransport) resolveUsingResolver(
...
-		var in *dns.Msg
-		if in, _, err = dnsClient.Exchange(ctx, msg, proto, resolver); err == nil {
-			for _, answer := range in.Answer {
-				if dns.RRToType(answer) == rrType {
-					switch rrType {
-					case dns.TypeA:
-						ips = append(ips, answer.(*dns.A).A.Addr.AsSlice())
-					case dns.TypeAAAA:
-						ips = append(ips, answer.(*dns.AAAA).AAAA.Addr.AsSlice())
+		if msg != nil {
```
welwood08: This could probably be simplified to `if msg == nil` -> `continue` instead. Either way, agree this needs a nil check of some kind.

ACEx86:
None of the suggested panic fixes is by AI i just came accross some of the program crashes that had to do with the deferer.

dnscrypt-proxy/resolve.go
```diff
@@ -24,6 +24,9 @@ func resolveQuery(server string, qName string, qType uint16, sendClientSubnet bo
 	transport.ReadTimeout = 2 * time.Second
 	client := &dns.Client{Transport: transport}
 	msg := dns.NewMsg(qName, qType)
+	if msg == nil {
+		return nil, errors.New("Query message error.")
```
welwood08: As per the [dns.NewMsg docs](https://pkg.go.dev/codeberg.org/miekg/dns#NewMsg), nil is returned when the type is unknown so we should be able to say that in the error message. Otherwise agree a nil check is needed here.

welwood08:
> None of the suggested panic fixes is by AI i just came accross some of the program crashes that had to do with the deferer.

Interesting, I'm not sure how some of those values could become `nil`. Do you have logs to share?

I assume it's not as simple as checking for `nil` at the same time as checking for `err != nil`? If so then it's strange to me that the rest of the function executes successfully on a non-`nil` value and it somehow transforms into `nil` in the deferred function...

ACEx86:
> > None of the suggested panic fixes is by AI i just came accross some of the program crashes that had to do with the deferer.
> 
> Interesting, I'm not sure how some of those values could become `nil`. Do you have logs to share?
> 
> I assume it's not as simple as checking for `nil` at the same time as checking for `err != nil`? If so then it's strange to me that the rest of the function executes successfully on a non-`nil` value and it somehow transforms into `nil` in the deferred function...

If you close the thread or the handle what will happen?
It's just a shield unexpected case in most of the times it hasn't to do with program exec.

welwood08:
> If you close the thread or the handle what will happen? It's just a shield unexpected case in most of the times it hasn't to do with program exec.

I'm not sure what you mean by closing the thread or the handle in this context. If you mean some action external to the Go program, I don't see how that would set a value to `nil` before a deferred function can call `Close()` on it.

Are you actually seeing panics at all these places with the deferred close or is it just defensive programming?

dnscrypt-proxy/dnsutils.go
```diff
@@ -459,7 +459,11 @@ func _dnsExchange(
 		if err != nil {
 			return DNSExchangeResponse{err: err}
 		}
-		defer pc.Close()
+		defer func(pc net.Conn) {
+			if pc != nil {
```
welwood08, line 463: The only possible way I can see this being `nil` here is if `net.DialTimeout` returned `nil, nil`. If we don't trust `net.DialTimeout` then it's simpler to check at the same time as `if err != nil` above.

dnscrypt-proxy/dnsutils.go
```diff
@@ -498,7 +502,11 @@ func _dnsExchange(
...
```
welwood08, line 506: This has more potential for an unknown proxy's `Dial` function to incorrectly return `nil, nil`. If that's the case, again the simpler guard is to check for nil at the same time as `if err != nil` above.

dnscrypt-proxy/local-doh.go
```diff
@@ -98,7 +98,11 @@ func (handler localDoHHandler) ServeHTTP(writer http.ResponseWriter, request *ht
 }
 
 func (proxy *Proxy) localDoHListener(acceptPc *net.TCPListener) {
-	defer acceptPc.Close()
+	defer func(acceptPc *net.TCPListener) {
+		if acceptPc != nil {
```
welwood08, line 102: In this case, if for some reason `acceptPc` is `nil` then there's no point doing anything in `localDoHListener`. The simpler solution here is an early `return`.

dnscrypt-proxy/sources.go
```diff
@@ -103,11 +103,19 @@ func writeSource(f string, bin, sig []byte) error {
 	if fSrc, err = safefile.Create(f, 0o644); err != nil {
 		return err
 	}
-	defer fSrc.Close()
+	defer func(fSrc *safefile.File) {
+		if fSrc != nil {
```
welwood08, line 107: If `safefile.Create` returns `nil, nil` for whatever reason, it is again simpler to check before the `defer` call and return some new error instead of continuing to attempt to write anything. I don't see any other code that would change the value of either `fSrc` or `fSig` after each `safefile.Create` call.

dnscrypt-proxy/systemd_linux.go
```diff
@@ -23,7 +24,11 @@ func (proxy *Proxy) addSystemDListeners() error {
 		proxy.listenAddresses = make([]string, 0)
 	}
 	for i, file := range files {
-		defer file.Close()
+		defer func(file *os.File) {
+			if file != nil {
```
welwood08, line 28: Once again, nothing seems to set the value of `file` to `nil` within the loop. If it's `nil` at the start of the loop for whatever reason, we can just `continue` before the `defer` instead of trying to register a listener at all.

ACEx86: The 'defined' execution isn't the only specified thing that can happen.
I don't know if this will result to a panic on linux but if you rm the file after use and before return what will happen trying to close without nil check? And i mean rm it from the system not the program.

welwood08: `file` here is unrelated to any Linux file handle, it will not magically become `nil` without something within the Go program setting it so.

jedisct1:
Barring the check for `dns.NewMsg` (thanks for catching that!), these changes are unnecessary and would actually hide problems.

If a function such as `safefile.Create` ever returns a `nil` value with the error also set to `nil`, that would be completely unexpected, and panicking is the correct thing to do.

PR #3073, opened 2025-12-29
Update plainNXTestPacket to return pointer to dns.Msg

ACEx86:
Change return type of plainNXTestPacket to pointer.

Return copies the lock value: type 'dns.Msg' contains 'atomic.Bool' contains 'interface{}' which is 'sync.Locker'


jedisct1:
?

ACEx86:
IDE warning. return *msg returns a copy of dns.Msg that in the struct has Mutex that is not to be copied?

jedisct1:
Thanks!

PR #3074, opened 2025-12-29
Refactor body_hash query parameter handling

ACEx86:
Fix a query leak.
Querys on some failures may be added together.

### DIFF
--- a/dnscrypt-proxy/xtransport.go
+++ b/dnscrypt-proxy/xtransport.go
@@ -682,7 +682,11 @@ func (xTransport *XTransport) Fetch(
 	if body != nil {
 		h := sha512.Sum512(*body)
 		qs := url.Query()
-		qs.Add("body_hash", hex.EncodeToString(h[:32]))
+		if qs.Has("body_hash") {
+			qs.Set("body_hash", hex.EncodeToString(h[:32]))
+		} else {
+			qs.Add("body_hash", hex.EncodeToString(h[:32]))
+		}
 		url2 := *url
 		url2.RawQuery = qs.Encode()
 		url = &url2


jedisct1:
I don't understand. Can you clarify what "querys on some failures" means?

How to reproduce that problem?

dnscrypt-proxy/xtransport.go
```diff
@@ -682,7 +682,11 @@ func (xTransport *XTransport) Fetch(
 	if body != nil {
 		h := sha512.Sum512(*body)
 		qs := url.Query()
-		qs.Add("body_hash", hex.EncodeToString(h[:32]))
+		if qs.Has("body_hash") {
+			qs.Set("body_hash", hex.EncodeToString(h[:32]))
+		} else {
+			qs.Add("body_hash", hex.EncodeToString(h[:32]))
+		}
```
welwood08, line 689: Why is increased complexity needed here? If `qs.Add` is problematic when the key already exists, why can't we just always use `qs.Set`?

ACEx86: // Set sets the key to value. It replaces any existing
// values.
func (v Values) Set(key, value string) {
	v[key] = []string{value}
}

// Add adds the value to key. It appends to any existing
// values associated with key.
func (v Values) Add(key, value string) {
	v[key] = append(v[key], value)
}

// The append built-in function appends elements to the end of a slice. If
// it has sufficient capacity, the destination is resliced to accommodate the
// new elements. If it does not, a new underlying array will be allocated.
// Append returns the updated slice. It is therefore necessary to store the
// result of append, often in the variable holding the slice itself:
//
//	slice = append(slice, elem1, elem2)
//	slice = append(slice, anotherSlice...)
//
// As a special case, it is legal to append a string to a byte slice, like this:
//
//	slice = append([]byte("hello "), "world"...)
func append(slice []Type, elems ...Type) []Type

welwood08: Thank you, that doesn't answer my question though. If the `body_hash` key doesn't exist, `append(nil, value)` is the same as `[]string{value}`. Why is the `if` check necessary when we could just always use `qs.Set` instead of `qs.Add`?

PR #3111, opened 2026-02-01
Update monitoring_ui.go: fix metric format

agvol:
Fix metric format


jedisct1:
Can you clarify why adding commas is necessary?

agvol:
> Can you clarify why adding commas is necessary?

https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf
labels = "{" [label *(COMMA label)] "}"

also, some scrapers got error, I check vmagent and telegraf

jedisct1:
Thank you!

PR #3113, opened 2026-02-02
Add configurable TLS version settings with smart defaults for DoH

icebluey:
This commit introduces comprehensive TLS version configuration for DNS-over-HTTPS connections with intelligent defaults and proper validation.

Features:
- Add tls_min_ver and tls_max_ver configuration options (support '1.2' and '1.3')
- Use pointer types (*string) to distinguish between unset (nil) and explicitly set values
- Smart default: When only tls_max_ver='1.2' is set, auto-adjust tls_min_ver to '1.2'
- Strict validation: Explicit conflicts (e.g., min=1.3, max=1.2) produce clear errors
- Default behavior: TLS 1.3 only when neither option is configured

TLS version handling in XTransport:
- Apply MinVersion and MaxVersion to TLS client configuration
- Handle tls_prefer_rsa interaction: warn and ignore when TLS 1.3 is required
- RSA cipher suites only available in TLS 1.2, automatically disabled for TLS 1.3

Configuration examples:
  tls_max_ver = '1.2'              -> Auto-adjusts to 1.2-1.2 (enables tls_prefer_rsa)
  tls_min_ver = '1.2'              -> 1.2-1.3 (allows both, enables tls_prefer_rsa)
  tls_min_ver = '1.3', max='1.3'   -> 1.3-1.3 (default, ignores tls_prefer_rsa)
  tls_min_ver = '1.3', max='1.2'   -> ERROR (explicit conflict)


jedisct1:
No.

jedisct1:
This is too complicated to understand, and it's difficult to justify usage of TLS 1.2 when servers support TLS 1.3.

The TLS 1.2 downgrade of RSA is a horrible workaround that was necessary after Go removed support for setting the cipher suite with TLS 1.3. But this is not something to capitalize on.

`dnscrypt-proxy` has too many knobs already. Adding new ones that virtually no one are going to use, and that can only degrade security and performance, is not a good idea.

PR #3117, opened 2026-02-08
support $RESOLVCONF keyword in forwarding rules

nlintn:
This adds the `$RESOLVCONF:<file>` keyword/function, which forwards the request to the resolvers specified in `<file>` (using [_resolv.conf_ syntax](https://en.wikipedia.org/wiki/Resolv.conf)).

Use cases include:

- Forwarding to the system resolvers in `/etc/resolv.conf` (provided your system uses this file)
- Forwarding to the DHCP DNS servers by specifying a file that is dynamically created and updated by your network configuration software (e.g. `/var/run/NetworkManager/resolv.conf`, created by NetworkManager, or `/run/systemd/resolve/resolv.conf`, created by systemd-networkd); in comparison to the `$DHCP` keyword this doesn't require root privileges on Linux and we don't have to send DHCP messages ourselves


jedisct1:
The format of `resolv.conf` files is not standardized and every operating system has their own features. I haven't looked at the dnsconf package, but it's likely that it won't be able to parse or properly handle an existing file on all operating systems.

Also, how does it work on Windows?

nlintn:
As the resolvers of all of `glibc`, `BSD libc` and `musl libc` use a `resolv.conf` file (defaulting to `/etc/resolv.conf`) with the syntax expected by the dnsconf package, it should work on almost all Unix(-like) systems (and probably all that keep their DNS config in a file, so most Linux distros, MacOS, FreeBSD etc.).
To be honest, the dnsconf package isn't really necessary here, as the syntax isn't *that* complicated (one option per line, e.g. nameservers specified by `nameserver <address>`). As we only care about the nameservers, writing a "parser" to extract those would be really simple and the commit would only consist of changes to two files instead of 16.

So Windows doesn't have a file containing the system resolvers (probably uses it's registry for that). As windows generates a `/etc/resolv.conf` file for each WSL instance, one could use those instead.
Other use cases would be a Linux file server sharing it's `resolv.conf` file and using that on windows, or just manually editing such a file (though that probably makes little sense).

nlintn:
Thank you! :)

PR #3118, opened 2026-02-09
Fix padding string generation in dnsutils.go

ACEx86:


jedisct1:
That would double the expected padding size by adding two "0" characters every time. What we want is add a single 0 byte (not the character).

Current code is fine.

ACEx86:
Update to the latest dns lib it doesn't accept 0x00 it will be corrupted with invalid encoding hex U+0000

ACEx86:
type PADDING struct {
	Padding string `dns:"hex"`
}

ACEx86:
So string.repeat 00 is the correct for latest update?

jedisct1:
Oh, indeed with the current version, this is hex. Good call. What a weird API.

PR #3216, opened 2026-05-23
Fix unsafe type assertions in ExtractClientIPStr

### LINKED ISSUE #3206: Unsafe type assertions in ExtractClientIPStr (common.go)
## Bug Description

The function `ExtractClientIPStr` in `common.go` uses unsafe type assertions (`(*net.UDPAddr)` and `(*net.TCPAddr)`) without the comma-ok idiom. If the `clientProto` label does not match the actual concrete type behind the `net.Addr` interface, the function **panics**.

### Affected Code

`common.go:182-195`:
```go
func ExtractClientIPStr(pluginsState *PluginsState) (string, bool) {
    if pluginsState.clientAddr == nil {
        return "", false
    }
    switch pluginsState.clientProto {
    case "udp":
        return (*pluginsState.clientAddr).(*net.UDPAddr).IP.String(), true  // UNSAFE
    case "tcp", "local_doh":
        return (*pluginsState.clientAddr).(*net.TCPAddr).IP.String(), true  // UNSAFE
    default:
        return "", false
    }
}
```

This is called from multiple plugin eval paths (block, allow, NX, query log) during query processing.

### Impact

A crafted query or internal state inconsistency could cause a type assertion panic, crashing the entire proxy. Note that `monitoring_ui.go` correctly uses the safe form with comma-ok.

### Fix

Use the comma-ok pattern for type assertions:
```go
case "udp":
    if udpAddr, ok := (*pluginsState.clientAddr).(*net.UDPAddr); ok && udpAddr != nil {
        return udpAddr.IP.String(), true
    }
case "tcp", "local_doh":
    if tcpAddr, ok := (*pluginsState.clientAddr).(*net.TCPAddr); ok && tcpAddr != nil {
        return tcpAddr.IP.String(), true
    }
```

hklcf:
Fixes #3206

Use comma-ok type assertion pattern to prevent panic if clientProto label mismatches the concrete net.Addr type.

### DIFF
--- a/dnscrypt-proxy/common.go
+++ b/dnscrypt-proxy/common.go
@@ -186,12 +186,15 @@ func ExtractClientIPStr(pluginsState *PluginsState) (string, bool) {
 	}
 	switch pluginsState.clientProto {
 	case "udp":
-		return (*pluginsState.clientAddr).(*net.UDPAddr).IP.String(), true
+		if udpAddr, ok := (*pluginsState.clientAddr).(*net.UDPAddr); ok && udpAddr != nil {
+			return udpAddr.IP.String(), true
+		}
 	case "tcp", "local_doh":
-		return (*pluginsState.clientAddr).(*net.TCPAddr).IP.String(), true
-	default:
-		return "", false
+		if tcpAddr, ok := (*pluginsState.clientAddr).(*net.TCPAddr); ok && tcpAddr != nil {
+			return tcpAddr.IP.String(), true
+		}
 	}
+	return "", false
 }
 
 // ExtractClientIPStrEncrypted extracts and optionally encrypts client IP string


jedisct1 (commented):
I do not think this should merge as-is.

The code change in `dnscrypt-proxy/common.go:188` and `dnscrypt-proxy/common.go:190` is mechanically safe for `ExtractClientIPStr` itself: a mismatched `net.Addr` would now return `ok=false` instead of panicking. The problem is that the PR and linked issue frame this as preventing a crafted-query crash, but I could not find a path where a DNS client controls either `clientProto` or the concrete `net.Addr`. The ingress sites set them together: UDP uses the `ReadFrom` result with `clientProto="udp"` in `dnscrypt-proxy/proxy.go:441` and `dnscrypt-proxy/proxy.go:462`, TCP uses `RemoteAddr` with `clientProto="tcp"` in `dnscrypt-proxy/proxy.go:492`, and local DoH builds a TCP address in `dnscrypt-proxy/local-doh.go:59` before using `"local_doh"`. So this is defensive programming against internal misuse, not a demonstrated remote crash fix.

That distinction matters because if the project wants to tolerate corrupted or mismatched internal state, this PR only handles one helper. Other nearby paths still assume the same invariant, for example `processIncomingQuery` dereferences `clientAddr` for logging in `dnscrypt-proxy/proxy.go:779` and UDP responses type-assert the connection/address path in `dnscrypt-proxy/query_processing.go:326`. Returning `ok=false` in `ExtractClientIPStr` would make some plugin logs silently skip the client IP, but it would not make the broader request path robust to the inconsistent state described in the PR.

There is also no regression test for the actual behavior being changed. `dnscrypt-proxy/common_test.go:8` covers nil, valid UDP, valid TCP, and unknown protocol, but it does not cover `"udp"` with a `*net.TCPAddr`, `"tcp"` with a `*net.UDPAddr`, or a non-nil `*net.Addr` containing a nil interface. Those are the cases that justify this patch, and they are exactly the cases most likely to regress later because the production paths normally preserve the invariant.

My recommendation is not to merge this PR as-is. If the maintainers want this small hardening change, I would reframe it as defensive cleanup rather than a security fix, add the missing regression tests, and avoid implying that it closes an externally triggerable crash unless a concrete production path is shown.

PR #3217, opened 2026-05-23
Add bounds check to HasTCFlag

### LINKED ISSUE #3208: HasTCFlag lacks bounds check (dnsutils.go)
## Bug Description

The `HasTCFlag` function in `dnsutils.go` accesses `packet[2]` without checking if the slice is at least 3 bytes long, which can cause a panic.

### Affected Code

`dnsutils.go:114-116`:
```go
func HasTCFlag(packet []byte) bool {
    return packet[2]&2 == 2  // out of bounds if len(packet) < 3
}
```

### Impact

All current callers guard against short packets before calling this function, but if a new caller is added in the future without such a guard, this would cause a panic and crash the proxy.

### Fix

Add a bounds check:
```go
func HasTCFlag(packet []byte) bool {
    if len(packet) < 3 {
        return false
    }
    return packet[2]&2 == 2
}
```

hklcf:
Fixes #3208

Prevent panic on packets shorter than 3 bytes.

### DIFF
--- a/dnscrypt-proxy/dnsutils.go
+++ b/dnscrypt-proxy/dnsutils.go
@@ -112,6 +112,9 @@ func RefusedResponseFromMessage(srcMsg *dns.Msg, refusedCode bool, ipv4 net.IP, 
 }
 
 func HasTCFlag(packet []byte) bool {
+	if len(packet) < 3 {
+		return false
+	}
 	return packet[2]&2 == 2
 }
 


jedisct1 (changes requested):
Thanks for the PR. I do not think this should merge as-is, and my recommendation is to close it rather than add this guard.

The proposed change in `dnscrypt-proxy/dnsutils.go:114` makes `HasTCFlag` silently return false for packets shorter than the DNS header field it reads. In the current codebase, though, this helper is only called after packet length has already been validated. `ApplyResponsePlugins` checks `len(packet) >= MinDNSPacketSize` before calling it on the unpack-error path at `dnscrypt-proxy/plugins.go:342`, and `sendResponse` returns before reaching the flag check when `len(response) < MinDNSPacketSize` at `dnscrypt-proxy/query_processing.go:306`.

Adding this check would also make `HasTCFlag` inconsistent with the adjacent raw-packet helpers in `dnscrypt-proxy/dnsutils.go`, such as `TransactionID`, `SetTransactionID`, and `Rcode`, which all assume they are operating on a validated DNS packet. That convention is useful here because malformed packets should be rejected at the boundary, not silently interpreted as a non-truncated response. If a future caller needs to handle untrusted bytes, it should perform the same `MinDNSPacketSize` validation as the existing callers before using these header helpers.

I also noticed the linked issue was closed by a maintainer as not planned, and there is no PR discussion or CI result indicating a currently failing path this fixes. For maintainability, I would prefer not to widen this helper's contract without a real call site or test demonstrating the need.

This is an automated review by Swival (https://swival.dev) using only open-source models, intended to help both the author and maintainers. My merge recommendation is not to merge this PR.

jedisct1:
Closing this since the linked issue was already closed as not planned, and the current callers already validate packet length before using `HasTCFlag`. The detailed rationale is in the review.

PR #3218, opened 2026-05-23
Fix missing defer cancel() in resolveQuery

### LINKED ISSUE #3209: Missing defer cancel() in resolveQuery (resolve.go)
## Bug Description

In `context_cancel.go`, the `cancel()` function from `context.WithTimeout` is not deferred, so if `client.Exchange` panics (e.g., due to a nil pointer in the DNS library), the context resources (timer goroutine) will leak until the timeout fires.

### Affected Code

`resolve.go:59-61`:
```go
ctx, cancel := context.WithTimeout(context.Background(), readTimeout)
response, rtt, err := client.Exchange(ctx, msg, "udp", server)
cancel()
```

Note that other parts of the codebase (e.g., `xtransport.go:483`, `plugin_dns64.go:205`) correctly use `defer cancel()`.

### Impact

If `client.Exchange` panics, `cancel()` is never called, causing a context/timer leak. While rare, this is a correctness issue.

### Fix

```go
ctx, cancel := context.WithTimeout(context.Background(), readTimeout)
defer cancel()
response, rtt, err := client.Exchange(ctx, msg, "udp", server)
```

hklcf:
Fixes #3209

Prevent context leak on panic by using defer for context cancellation.

### DIFF
--- a/dnscrypt-proxy/resolve.go
+++ b/dnscrypt-proxy/resolve.go
@@ -57,8 +57,8 @@ func resolveQuery(server string, qName string, qType uint16, sendClientSubnet bo
 		msg.ID = dns.ID()
 		msg.Data = nil // Clear packed data so Exchange will re-pack with new ID
 		ctx, cancel := context.WithTimeout(context.Background(), readTimeout)
+		defer cancel()
 		response, rtt, err := client.Exchange(ctx, msg, "udp", server)
-		cancel()
 		if neterr, ok := err.(net.Error); ok && neterr.Timeout() {
 			readTimeout *= 2
 			continue


jedisct1 (changes requested):
Thanks for sending this. I reviewed the PR, the linked issue, the surrounding `resolveQuery` implementation, the other local `context.WithTimeout` call sites, and the current PR discussion. There are no existing review comments or CI checks on this PR, and the linked issue was already closed as not planned.

I do not recommend merging this change. The modified code in `dnscrypt-proxy/resolve.go:59` creates a timeout context inside a retry loop and currently calls `cancel()` immediately after `client.Exchange` returns at `dnscrypt-proxy/resolve.go:61`. That is the right lifetime for this context: each retry gets its own timeout, and the timer is released as soon as that specific exchange is done.

Changing that call to `defer cancel()` inside the loop does not make the normal path safer. On success or on a non-timeout error, the defer runs when `resolveQuery` returns, but on the timeout retry path it postpones cleanup until after the loop exits instead of doing it immediately at the end of that attempt. There are only three attempts here, so this is not a large leak, but it moves the code away from the precise lifetime management used in the current implementation without adding a real benefit.

The stated justification is a panic in `client.Exchange`. I do not think that is a good reason to change this code. The repository does not recover panics on this path, so such a panic would unwind out of `resolveQuery` and likely terminate the process rather than leave a long-running operation behind. Also, the timeout is already short and bounded by `readTimeout`. This makes the proposed fix disproportionate to the claimed problem, and it introduces a less direct cancellation pattern in a retry loop.

The existing codebase uses `defer cancel()` where the context lifetime naturally matches the whole function scope, for example `dnscrypt-proxy/plugin_dns64.go:204` and `dnscrypt-proxy/xtransport.go:483`. In contrast, loop-scoped exchange attempts such as `dnscrypt-proxy/plugin_forward.go:382` cancel explicitly before continuing. The current `resolveQuery` code follows that latter pattern, and this PR would make it less consistent with the surrounding retry logic.

For that reason, I would close this PR rather than keep iterating on it. If there is a real panic from the DNS library to handle, the better fix would be to identify and address that panic directly, or add an intentional recovery boundary with tests if the project wants one. This one-line defer change does not meaningfully improve correctness and is not necessary for maintainability.

PR #3219, opened 2026-05-23
Fix double strings.ToLower in ipcrypt.go

### LINKED ISSUE #3215: Double strings.ToLower in ipcrypt.go
## Bug Description

In `ipcrypt.go`, the `algorithm` parameter is converted to lowercase twice: once on line 44 and again on line 48 inside the switch statement. While harmless, this is redundant and confusing.

### Affected Code

`ipcrypt.go:44` and `ipcrypt.go:48`:
```go
config := &IPCryptConfig{
    Algorithm: strings.ToLower(algorithm),  // first conversion
}
switch strings.ToLower(algorithm) {  // second conversion
```

### Fix

Use the already-normalized value from the config struct in the switch:
```go
switch config.Algorithm {
```

hklcf:
Fixes #3215

Use already-normalized config.Algorithm value in the switch statement instead of calling strings.ToLower a second time.

### DIFF
--- a/dnscrypt-proxy/ipcrypt.go
+++ b/dnscrypt-proxy/ipcrypt.go
@@ -45,7 +45,7 @@ func NewIPCryptConfig(keyHex string, algorithm string) (*IPCryptConfig, error) {
 	}
 
 	// Validate key length and prepare config based on algorithm
-	switch strings.ToLower(algorithm) {
+	switch config.Algorithm {
 	case "ipcrypt-deterministic":
 		// Deterministic IPCrypt uses 16-byte keys
 		if len(key) != 16 {


jedisct1 (commented):
Thanks for the contribution. I looked through the PR metadata, the linked issue, the current diff, and the surrounding `NewIPCryptConfig` implementation in `dnscrypt-proxy/ipcrypt.go`. I also checked that `go test -run TestNewIPCryptConfig ./dnscrypt-proxy` still passes on the PR branch.

I do not recommend merging this PR as it stands. The code change at `dnscrypt-proxy/ipcrypt.go:48` is behavior-preserving because `config.Algorithm` is assigned from `strings.ToLower(algorithm)` just a few lines earlier, so replacing `switch strings.ToLower(algorithm)` with `switch config.Algorithm` does not fix a correctness issue or remove a meaningful maintenance burden.

The linked issue was also automatically closed as not following the project issue process, and the reported problem is a harmless duplicate call rather than a user-visible bug. If this area is cleaned up later, I think a more complete and intentional change would normalize the algorithm once near the start of `NewIPCryptConfig`, before the `"none"` early return, and then use that normalized value consistently. This PR only changes the switch and leaves the rest of the normalization flow unchanged.

Since the change is tiny but not necessary or justified enough to carry as a separate PR, my recommendation is to close this rather than merge it. This is an automated review by Swival (https://swival.dev) using only open-source models, intended to help both the PR author and the maintainers.

jedisct1:
Closing this because the proposed change is only a behavior-preserving cleanup for a duplicate `strings.ToLower()` call, and the linked issue was already closed as invalid. If we revisit this code, it would be better to normalize the algorithm once at the top of `NewIPCryptConfig` and use that value consistently rather than merge this narrow cleanup on its own.

PR #3220, opened 2026-05-23
Fix log level config silently overridden by env var check

### LINKED ISSUE #3211: Log level config silently overridden by env var check (config_loader.go)
## Bug Description

In `config_loader.go`, the log level set by the user in the configuration file is overridden by an environment variable check. If a user explicitly sets `log_level = 0` (debug) in the config file but does not set the `DEBUG` environment variable, the log level gets silently overridden to info.

### Affected Code

`config_loader.go:32-34`:
```go
if dlog.LogLevel() <= dlog.SeverityDebug && os.Getenv("DEBUG") == "" {
    dlog.SetLogLevel(dlog.SeverityInfo)
}
```

### Impact

User-configured debug logging is silently ignored. Users who set `log_level = 0` in their config file expecting debug output will not see it unless they also set the `DEBUG` environment variable. This is surprising behavior and contradicts the explicit configuration file setting.

### Fix

Either remove this override entirely (letting the config file control log level) or document it clearly and only apply it when the user hasn't explicitly configured the log level:
```go
// Only override if the user hasn't explicitly set a log level in the config
if os.Getenv("DEBUG") == "" && defaultLogLevelUsed {
    // ...
}
```

hklcf:
Fixes #3211

Remove the DEBUG env var override that silently promoted user-configured debug logging to info.

### DIFF
--- a/dnscrypt-proxy/config_loader.go
+++ b/dnscrypt-proxy/config_loader.go
@@ -29,9 +29,6 @@ func configureLogging(proxy *Proxy, flags *ConfigFlags, config *Config) {
 	if config.LogLevel >= 0 && config.LogLevel < int(dlog.SeverityLast) {
 		dlog.SetLogLevel(dlog.Severity(config.LogLevel))
 	}
-	if dlog.LogLevel() <= dlog.SeverityDebug && os.Getenv("DEBUG") == "" {
-		dlog.SetLogLevel(dlog.SeverityInfo)
-	}
 	dlog.TruncateLogFile(config.LogFileLatest)
 
 	isCommandMode := false


jedisct1 (approved):
Thanks for the patch. I reviewed PR #3220 against `ConfigLoad()` and `configureLogging()`, the linked issue #3211, the current example configuration, and the visible pull request discussion. There were no PR comments, review threads, or replies to account for.

My recommendation is to merge this as is, subject to the normal maintainer checks. The change in `dnscrypt-proxy/config_loader.go` removes the block that runs immediately after applying `config.LogLevel` and downgrades `SeverityDebug` to `SeverityInfo` whenever `DEBUG` is unset. That matches the reported issue: with `log_level = 0`, the explicit TOML value is accepted at `dnscrypt-proxy/config_loader.go:29` and then overwritten at `dnscrypt-proxy/config_loader.go:32`. The sample config documents 0 as a valid log level at `dnscrypt-proxy/example-dnscrypt-proxy.toml:220`, so requiring a separate `DEBUG` environment variable makes the documented setting misleading.

I also checked the surrounding load flow in `dnscrypt-proxy/config.go:348` and `dnscrypt-proxy/config.go:383`. `newConfig()` seeds `LogLevel` from the initial dlog level, TOML decoding can replace it, and `configureLogging()` is the place where the effective value is applied. Removing the extra `DEBUG` gate is the smallest clear fix and does not disturb command mode handling, syslog/file logging, or descriptor inheritance. The `os` import in `config_loader.go` remains needed by other code in the same file, so this deletion does not leave cleanup behind.

I did not find a correctness regression from this change. It intentionally allows configured debug logging to work, while the default still stays at notice through `dlog.Init()` and `newConfig()` when no `log_level` is specified. I ran `go test ./dnscrypt-proxy`, and it passed.

This is an automated review by Swival (https://swival.dev) using only open-source models. The goal is to help both the pull request author and the maintainers make the safest decision here.

jedisct1:
This is intentional. Debug level is for maintainers.

Prior to this we got a lot of reports from users who freaked out about log debug-only, internal messages they didn't understand.

n-thumann:
+1 for this change or at least adding a hint to the documentation. Just like https://github.com/DNSCrypt/dnscrypt-proxy/pull/1370, I spent too much time trying to figure out why debug logs weren't working, even though I set `log_level = 0` 😕

PR #3221, opened 2026-05-23
Fix data race on source.refresh field in sources.go

### LINKED ISSUE #3204: Data race on source.refresh field in sources.go
## Bug Description

There is a data race on the `source.refresh` field in `sources.go`. The `time.Time` value is written at lines 180, 185, and 211 without holding the source lock, while `PrefetchSources()` reads it at line 267 without any synchronization.

### Affected Code

`sources.go:180` (in `fetchWithCache`):
```go
source.refresh = now.Add(ttl)  // write without lock
```

`sources.go:267` (in `PrefetchSources`):
```go
if source.refresh.IsZero() || source.refresh.After(now) {  // read without lock
    continue
}
```

### Impact

`time.Time` is a 16-byte struct on 64-bit platforms. Concurrent write and read without synchronization is undefined behavior in Go's memory model. A partially written value could cause `IsZero()` or `After()` to read corrupted bytes, leading to incorrect prefetch scheduling or missed source updates.

### Fix

Wrap writes in the source lock:
```go
source.Lock()
source.refresh = now.Add(ttl)
source.Unlock()
```

And reads:
```go
source.RLock()
needsRefresh := source.refresh.IsZero() || !source.refresh.After(now)
source.RUnlock()
```

hklcf:
Fixes #3204

Wrap writes and reads to source.refresh with proper locking to prevent data race between fetchWithCache and PrefetchSources.

### DIFF
--- a/dnscrypt-proxy/sources.go
+++ b/dnscrypt-proxy/sources.go
@@ -177,12 +177,16 @@ func (source *Source) fetchWithCache(xTransport *XTransport) (time.Duration, err
 		return 0, err
 	}
 	if ttl > 0 {
+		source.Lock()
 		source.refresh = now.Add(ttl)
+		source.Unlock()
 		return 0, err
 	}
 
 	ttl = MinimumPrefetchInterval
+	source.Lock()
 	source.refresh = now.Add(ttl)
+	source.Unlock()
 	var bin, sig []byte
 	for _, srcURL := range source.urls {
 		dlog.Infof("Source [%s] loading from URL [%s]", source.name, srcURL)
@@ -208,7 +212,9 @@ func (source *Source) fetchWithCache(xTransport *XTransport) (time.Duration, err
 	}
 	source.updateCache(bin, sig)
 	ttl = source.prefetchDelay
+	source.Lock()
 	source.refresh = now.Add(ttl)
+	source.Unlock()
 	return ttl, nil
 }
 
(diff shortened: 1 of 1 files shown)


jedisct1 (commented):
Thanks for taking a look at the race around `source.refresh`. I agree with the intent, but I would not merge this version as-is because the change alters the meaning of a zero `refresh` value in `PrefetchSources()`.

The original code in `dnscrypt-proxy/sources.go` skipped sources when `source.refresh.IsZero()` was true. That is not just an uninitialized case. `fetchWithCache()` returns at `dnscrypt-proxy/sources.go:176` when a source has no URLs, before assigning `source.refresh`, and URL-less sources are an explicitly documented configuration mode in `dnscrypt-proxy/example-dnscrypt-proxy.toml:753` where the cache file and signature must already be present. With the new predicate at `dnscrypt-proxy/sources.go:274`, a zero refresh is treated as needing refresh, so those cache-only sources will be retried by the prefetch loop every `MinimumPrefetchInterval` even though there is no remote URL to fetch. That is a behavioral regression and can produce unnecessary repeated cache reads and update attempts for a supported setup.

The locking itself is the right direction, but the predicate needs to preserve the old behavior while reading under `RLock`, for example by computing whether to skip from `source.refresh.IsZero() || source.refresh.After(now)` instead of folding zero into the refresh-needed case. It would also be worth adding a small regression test for a source with no URLs and a zero `refresh` value, since the current `TestPrefetchSources` setup forces `refresh` to `d.timeUpd` and does not cover this path.

I checked the PR metadata and visible discussion, and there were no existing comments or review threads to account for. I also ran `go test ./dnscrypt-proxy` and `go test -race ./dnscrypt-proxy -run TestPrefetchSources`; both pass, but they do not catch the cache-only source regression described above.

This is an automated review by Swival (https://swival.dev), using only open-source models, intended to help both the author and the maintainers. My recommendation is to close this PR and reopen or update it with the predicate fixed so the synchronization change is behavior-preserving.

jedisct1:
Closing based on the review above: the current patch changes the zero-refresh behavior for supported cache-only sources. Please feel free to reopen or send an updated PR that preserves the existing predicate while adding synchronization.

PR #3222, opened 2026-05-23
Fix cache update race in plugin_cloak.go

### LINKED ISSUE #3210: Race condition in plugin_cloak.go cache update
## Bug Description

The `Eval` method in `plugin_cloak.go` has a race condition when two concurrent queries need to resolve the same cloaked hostname. The "last writer wins" cache update pattern means that a concurrent resolution can overwrite another goroutine's results.

### Affected Code

`plugin_cloak.go:300-334`:
```go
plugin.RUnlock()  // line 300 - release read lock
// ... DNS resolution (takes time, no lock held)
plugin.Lock()     // line 319
cloakedName.ipv4 = foundIPs[:n]  // line 325 - write
plugin.Unlock()   // line 331
```

### Impact

Goroutine A resolves "example.com" to IPs [1.2.3.4, 5.6.7.8] and stores them. Goroutine B (concurrently) resolves the same name to [9.10.11.12, 13.14.15.16] and overwrites A's result. The cached IPs for the cloaked name can flap between two different resolution results depending on timing. This can cause inconsistent DNS responses for clients.

### Fix

Use a per-name mutex or a compare-and-swap-like pattern to serialize DNS resolution per cloaked name. Alternatively, skip the cache update if another goroutine has already updated it:
```go
plugin.Lock()
// Only update if another goroutine hasn't already done so
if cloakedName.lastUpdate4 == nil || now.After(*cloakedName.lastUpdate4) {
    cloakedName.lastUpdate4 = &now
    cloakedName.ipv4 = foundIPs[:n]
}
plugin.Unlock()
```

hklcf:
Fixes #3210

Only update cached IPs when they haven't been populated by another goroutine, preventing last-writer-wins race on concurrent cloaked name resolution.

### DIFF
--- a/dnscrypt-proxy/plugin_cloak.go
+++ b/dnscrypt-proxy/plugin_cloak.go
@@ -321,11 +321,15 @@ func (plugin *PluginCloak) Eval(pluginsState *PluginsState, msg *dns.Msg) error 
 			n := Min(16, len(foundIPs))
 			switch qtype {
 			case dns.TypeA:
-				cloakedName.lastUpdate4 = &now
-				cloakedName.ipv4 = foundIPs[:n]
+				if len(cloakedName.ipv4) == 0 {
+					cloakedName.lastUpdate4 = &now
+					cloakedName.ipv4 = foundIPs[:n]
+				}
 			case dns.TypeAAAA:
-				cloakedName.lastUpdate6 = &now
-				cloakedName.ipv6 = foundIPs[:n]
+				if len(cloakedName.ipv6) == 0 {
+					cloakedName.lastUpdate6 = &now
+					cloakedName.ipv6 = foundIPs[:n]
+				}
 			}
 		}
 		plugin.Unlock()


jedisct1 (commented):
This is an automated review by Swival (https://swival.dev) using only open-source models, intended to help both the PR author and the maintainers.

I do not recommend merging this PR as-is. The change is very small, but it breaks the existing refresh semantics for hostname-based cloaking rules.

In `dnscrypt-proxy/plugin_cloak.go`, `Eval` intentionally re-resolves a cloaked hostname when `lastUpdate4` or `lastUpdate6` has exceeded `cloak_ttl`. With the new `len(cloakedName.ipv4) == 0` and `len(cloakedName.ipv6) == 0` guards at `dnscrypt-proxy/plugin_cloak.go:324` and `dnscrypt-proxy/plugin_cloak.go:329`, an entry that was successfully resolved once will never have its cached IPs or `lastUpdate` timestamp refreshed again. Once the TTL expires, every matching query will still go through `resolveUsingServers`, but the result will be discarded because the slice is already non-empty, and the response will continue using the original stale addresses.

That is a user-visible regression for cloaking rules that point to hostnames rather than literal IPs. It also creates avoidable upstream lookups after expiration, because `lastUpdate` remains old forever and `expired` will stay true on every later query.

The issue description suggested avoiding a concurrent overwrite, but this implementation uses slice length as the guard, which cannot distinguish “another goroutine already refreshed this entry” from “this entry has old cached data that now needs to be replaced.” A safer fix would need to re-check the observed cache state under the write lock, for example by comparing the `lastUpdate` value that was seen before unlocking, or by serializing per-name resolution, while still allowing expired cached addresses to be replaced.

I also did not find PR discussion or review comments to account for this tradeoff, and there are no CI checks reported on the branch. I ran `go test -race ./...` locally and it passed, but that does not cover this stale-cache behavior.

Thanks for taking a look at the reported concurrency concern. I think this should be closed or reworked rather than merged in its current form, because it fixes the alleged last-writer behavior by disabling legitimate cache refreshes.

PR #3223, opened 2026-05-23
Fix data race on internalResolverReady field

### LINKED ISSUE #3205: Data race on internalResolverReady field in xtransport.go
## Bug Description

There is a data race on `proxy.xTransport.internalResolverReady` in `serversInfo.go`. Multiple goroutines (one per server, up to `certRefreshConcurrency`) concurrently write to the plain `bool` field without any synchronization.

### Affected Code

`serversInfo.go:264` (writer):
```go
go func(registeredServer *RegisteredServer) {
    err := serversInfo.refreshServer(proxy, registeredServer.name, registeredServer.stamp)
    if err == nil {
        proxy.xTransport.internalResolverReady = true  // RACE
    }
    errorChannel <- err
    <-countChannel
}(&registeredServers[i])
```

`xtransport.go:562` (reader):
```go
if xTransport.internalResolverReady {
```

### Impact

While bool writes are often atomic in practice, Go's memory model does not guarantee visibility across goroutines without synchronization. A stale `false` value could cause the proxy to skip internal resolvers and fall through to bootstrap resolvers unnecessarily, or a stale `true` value could cause the proxy to use internal resolvers that aren't actually ready yet.

### Fix

Use `atomic.Bool` from `sync/atomic`:
```go
// Change declaration to:
internalResolverReady atomic.Bool

// Writer:
proxy.xTransport.internalResolverReady.Store(true)

// Reader:
if proxy.xTransport.internalResolverReady.Load() {
```

hklcf:
Fixes #3205

Use atomic.Bool instead of plain bool to synchronize access across goroutines in server refresh loop.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -292,7 +292,7 @@ func (proxy *Proxy) StartProxy() {
 			dlog.Fatal(err)
 		}
 	}
-	proxy.xTransport.internalResolverReady = false
+	proxy.xTransport.internalResolverReady.Store(false)
 	proxy.xTransport.internalResolvers = proxy.listenAddresses
 	liveServers, err := proxy.serversInfo.refresh(proxy)
 	if liveServers > 0 {
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -261,7 +261,7 @@ func (serversInfo *ServersInfo) refresh(proxy *Proxy) (int, error) {
 		go func(registeredServer *RegisteredServer) {
 			err := serversInfo.refreshServer(proxy, registeredServer.name, registeredServer.stamp)
 			if err == nil {
-				proxy.xTransport.internalResolverReady = true
+				proxy.xTransport.internalResolverReady.Store(true)
 			}
 			errorChannel <- err
 			<-countChannel
--- a/dnscrypt-proxy/xtransport.go
+++ b/dnscrypt-proxy/xtransport.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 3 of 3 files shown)


jedisct1:
Thanks for the focused fix. I reviewed the diff against the surrounding startup, refresh, and resolver code paths, and I also checked the PR conversation and linked issue context. There are no existing PR comments or review threads to reconcile, and GitHub does not show any CI status for this PR.

The actual code change is small and matches the problem described in #3205. `serversInfo.refresh()` starts concurrent goroutines and writes `proxy.xTransport.internalResolverReady` when a server refresh succeeds, while `XTransport.resolve()` can read that same state from the resolution path. Converting the field to `atomic.Bool` in `dnscrypt-proxy/xtransport.go:82`, using `Store(false)` during startup in `dnscrypt-proxy/proxy.go:295`, using `Store(true)` after a successful refresh in `dnscrypt-proxy/serversInfo.go:264`, and using `Load()` in `dnscrypt-proxy/xtransport.go:563` removes the unsynchronized bool access without changing the intended behavior.

I did not find a correctness regression in this patch. The field is already internal to `XTransport`, the module targets a Go version where `atomic.Bool` is available, and the repository already uses both `sync/atomic` and typed atomics in nearby code such as `plugin_forward.go`. I also checked the other current references to `internalResolverReady` and did not find a remaining plain read or write.

I verified the branch with `go test ./dnscrypt-proxy` and `go test -race ./dnscrypt-proxy`, and both passed locally. My merge recommendation for the code itself is that this is safe to merge if the maintainers want to accept the contribution.

I am closing the PR as part of maintainer triage rather than because I found a blocking defect in the patch. This is an automated review by Swival (https://swival.dev) using only open-source models, intended to help both the pull request author and the maintainers.

PR #3224, opened 2026-05-23
Fix TOCTOU race in refreshServer

### LINKED ISSUE #3207: TOCTOU race in refreshServer (serversInfo.go)
## Bug Description

There is a TOCTOU (Time-of-Check-Time-of-Use) race condition in `refreshServer` in `serversInfo.go`. The initial check for whether a server exists in `serversInfo.inner` is done under `RLock`, but the actual insertion into the slice is done later under a separate `Lock`. Between these two lock acquisitions, another goroutine could add the same server, leading to duplicate entries.

### Affected Code

`serversInfo.go:207-244`:
```go
func (serversInfo *ServersInfo) refreshServer(proxy *Proxy, name string, stamp stamps.ServerStamp) error {
    serversInfo.RLock()
    isNew := true
    for _, oldServer := range serversInfo.inner {
        if oldServer.Name == name {
            isNew = false
            break
        }
    }
    serversInfo.RUnlock()
    // ... fetchServerInfo (no lock held, takes time) ...
    isNew = true
    serversInfo.Lock()
    // check again under write lock
    for i, oldServer := range serversInfo.inner {
        if oldServer.Name == name {
            serversInfo.inner[i] = &newServer
            isNew = false
            break
        }
    }
    serversInfo.Unlock()
    if isNew {
        serversInfo.Lock()
        serversInfo.inner = append(serversInfo.inner, &newServer)
        serversInfo.Unlock()
    }
}
```

### Impact

The race can be triggered when `refreshServer` is called concurrently from two paths:
1. The periodic cert refresh goroutine (`refresh()` → `refreshServer()`)
2. The ODoH key update path in query processing (`processODoHQuery()` → `refreshServer()`)

Both goroutines can pass the initial `RLock` check before either acquires the `Lock`, resulting in duplicate `ServerInfo` pointers in `serversInfo.inner`.

### Fix

Remove the first RLock check entirely and use only the second Lock-based check. The `isNew` variable can be determined solely from the write-locked section:
```go
func (serversInfo *ServersInfo) refreshServer(proxy *Proxy, name string, stamp stamps.ServerStamp) error {
    isNew := true
    serversInfo.RLock()
    for _, oldServer := range serversInfo.inner {
        if oldServer.Name == name {
            isNew = false
            break
        }
    }
    serversInfo.RUnlock()
    newServer, err := fetchServerInfo(proxy, name, stamp, isNew)
    if err != nil {
        return err
    }
    newServer.rtt = ewma.NewMovingAverage(RTTEwmaDecay)
    newServer.rtt.Set(float64(newServer.initialRtt))
    serversInfo.Lock()
    found := false
    for i, oldServer := range serversInfo.inner {
        if oldServer.Name == name {
            serversInfo.inner[i] = &newServer
            found = true
            break
        }
    }
    if !found {
        serversInfo.inner = append(serversInfo.inner, &newServer)
        proxy.serversInfo.registerServer(name, stamp)
    }
    serversInfo.Unlock()
    return nil
}
```

hklcf:
Fixes #3207

Combine the check-and-append into a single locked section to prevent duplicate server entries when refreshServer is called concurrently.

### DIFF
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -223,20 +223,20 @@ func (serversInfo *ServersInfo) refreshServer(proxy *Proxy, name string, stamp s
 	}
 	newServer.rtt = ewma.NewMovingAverage(RTTEwmaDecay)
 	newServer.rtt.Set(float64(newServer.initialRtt))
-	isNew = true
 	serversInfo.Lock()
+	found := false
 	for i, oldServer := range serversInfo.inner {
 		if oldServer.Name == name {
 			serversInfo.inner[i] = &newServer
-			isNew = false
+			found = true
 			break
 		}
 	}
-	serversInfo.Unlock()
-	if isNew {
-		serversInfo.Lock()
+	if !found {
 		serversInfo.inner = append(serversInfo.inner, &newServer)
-		serversInfo.Unlock()
+	}
+	serversInfo.Unlock()
+	if !found {
 		proxy.serversInfo.registerServer(name, stamp)
 	}
 


jedisct1 (commented):
I reviewed this against master, including the linked issue #3207, the single commit, the diff, and the visible PR/issue discussion. There are no visible comments, review threads, or checks reported for the branch.

My merge recommendation is: this looks safe to merge, though I would strongly prefer a small regression test if the maintainers want this race to stay fixed long term.

The change in `dnscrypt-proxy/serversInfo.go` does fix the specific duplicate-entry race it claims to address. In the current code, `refreshServer` checks for an existing `ServerInfo`, releases the write lock, and only later reacquires it to append. Two concurrent calls can both observe “not found” before either append happens, then both append the same server. The PR correctly makes the replace-or-append decision inside one locked section, so only one goroutine can append a missing server.

I do not see a correctness regression from moving `registerServer` after the unified locked section. `registerServer` is already idempotent, and calling it after releasing the lock avoids self-deadlocking on the same `ServersInfo` mutex. For the `!found` case, the server has already been inserted into `inner`, so a later call to `registerServer` only updates the registered-server list.

One thing this PR does not fix is the earlier `isNew` check before `fetchServerInfo`. Two concurrent first-time refreshes can still both call `fetchServerInfo(..., isNew=true)` before either one inserts the server. From the surrounding code, that appears to affect log severity rather than the correctness of the server list. So I would not block this PR on that, but it is worth being precise: this fixes duplicate insertion into `serversInfo.inner`, not all duplicated first-time refresh work.

I also noticed that the ODoH key update path still iterates `proxy.serversInfo.registeredServers` without taking `serversInfo.RLock()` in `dnscrypt-proxy/query_processing.go`. That appears to be pre-existing and outside the exact changed hunk, but it is relevant to the PR description because the description names `processODoHQuery()` as one of the concurrent paths. This PR does not make that worse, and it does not need to solve it to fix the duplicate `inner` append, but maintainers may want to track it separately if they are auditing this area for races.

The main maintainability gap is test coverage. There does not appear to be an existing `refreshServer` concurrency regression test, and this PR does not add one. Because `fetchServerInfo` is a hard dependency and performs network/protocol work, a clean test may require a small seam or a targeted helper around the replace-or-append block. I would not require that if the maintainers want the minimal production fix, but without a test this exact bug could be reintroduced during a future cleanup.

PR #3225, opened 2026-05-23
Fix crypto/rand.Read error silently suppressed

### LINKED ISSUE #3212: crypto/rand.Read error silently suppressed (serversInfo.go)
## Bug Description

In `serversInfo.go`, the error from `crypto/rand.Read` is silently discarded when generating random padding for test probe packets. If `crypto/rand.Read` fails (possible on certain platforms like containers with limited entropy), `paddingData` remains all zeros, weakening the DNS query size obfuscation.

### Affected Code

`serversInfo.go:867` and `serversInfo.go:888`:
```go
paddingData := make([]byte, 16)
_, _ = crypto_rand.Read(paddingData)  // error silently ignored
```

### Impact

Weakens EDNS0 padding protection. While `crypto/rand.Read` rarely fails on typical systems, silently suppressing errors in a security-critical DNS proxy violates defense-in-depth principles.

### Fix

Log the error:
```go
if _, err := crypto_rand.Read(paddingData); err != nil {
    dlog.Warnf("Failed to read random padding: %v", err)
}
```

hklcf:
Fixes #3212

Log errors from crypto/rand.Read when generating padding for test probe packets instead of silently discarding them.

### DIFF
--- a/dnscrypt-proxy/serversInfo.go
+++ b/dnscrypt-proxy/serversInfo.go
@@ -864,7 +864,9 @@ func dohTestPacket(msgID uint16) []byte {
 	msg.UDPSize = uint16(MaxDNSPacketSize)
 	msg.Security = false
 	paddingData := make([]byte, 16)
-	_, _ = crypto_rand.Read(paddingData)
+	if _, err := crypto_rand.Read(paddingData); err != nil {
+		dlog.Warnf("Failed to generate random padding: %v", err)
+	}
 	padding := &dns.PADDING{Padding: hex.EncodeToString(paddingData)}
 	msg.Pseudo = append(msg.Pseudo, padding)
 	if err := msg.Pack(); err != nil {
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks for the PR. I'm going to close this one.

The project currently requires Go 1.25 (`go.mod:3`), and with the supported Go API `crypto/rand.Read` always fills the buffer and never returns an error to the caller. If the underlying randomness source fails, Go terminates irrecoverably instead of letting this branch run. So the new warnings in `dnscrypt-proxy/serversInfo.go` are not reachable for supported builds.

There is also a correctness issue with the proposed behavior if we considered the older failure model: logging the error and then continuing would still send padding derived from the zero-filled buffer, which is the degraded behavior the linked issue is concerned about. That means this would make the problem visible in logs, but would not actually fix it.

Given that these packets are resolver probe packets, and the repository already uses deterministic padding bytes in other DNS padding paths, I don't think this change is necessary or proportionate. Closing this as not planned, but thank you for taking the time to look at it.

jedisct1:
Closing as not planned; see the review comment above for the reasoning.

PR #3226, opened 2026-05-23
Fix discarded error from serversInfo.refresh() in proxy.go

### LINKED ISSUE #3214: Discarded error from serversInfo.refresh() in proxy.go
## Bug Description

In the periodic server refresh goroutine in `proxy.go`, the error from `serversInfo.refresh()` is silently discarded using the blank identifier. This error contains information about which servers failed and why.

### Affected Code

`proxy.go:331`:
```go
liveServers, _ = proxy.serversInfo.refresh(proxy)
```

### Impact

Diagnostic information about server failures is lost. If all servers fail, `liveServers` will be 0, but the reason for the failure is not logged, making debugging difficult.

### Fix

Log the error:
```go
liveServers, err = proxy.serversInfo.refresh(proxy)
if err != nil {
    dlog.Warnf("Server refresh error: %v", err)
}
```

hklcf:
Fixes #3214

Log the error returned by serversInfo.refresh() to aid debugging when server certificate refresh fails.

### DIFF
--- a/dnscrypt-proxy/proxy.go
+++ b/dnscrypt-proxy/proxy.go
@@ -328,7 +328,10 @@ func (proxy *Proxy) StartProxy() {
 					delay = proxy.certRefreshDelayAfterFailure
 				}
 				clocksmith.Sleep(delay)
-				liveServers, _ = proxy.serversInfo.refresh(proxy)
+				liveServers, err = proxy.serversInfo.refresh(proxy)
+				if err != nil {
+					dlog.Warnf("Server refresh error: %v", err)
+				}
 				if liveServers > 0 {
 					proxy.certIgnoreTimestamp = false
 				}


jedisct1:
Thanks for taking a look at this.

I'm going to close this PR because I don't think this is the right fix for the diagnostic issue. `serversInfo.refresh()` currently returns only a lossy aggregate error: it keeps whichever worker error happens to be received last, and then clears the error completely if at least one server refreshed successfully. So logging that value from the periodic refresh loop would still miss partial failures, and when all servers fail it may point at an arbitrary resolver rather than the useful underlying cause.

There is also a log-noise concern here. When no servers are reachable, the retry delay is intentionally short, so this warning could repeat every 10 seconds without adding much more information than the resolver-specific logs already emitted lower in the refresh path.

If we want to improve this, the better place is probably inside the per-server refresh path where the server name and exact failure are still available, while being careful not to duplicate errors that are already logged by the lower-level DNSCrypt and DoH fetch code. Thanks again, but I don't want to merge this version as-is.

PR #3269, opened 2026-07-08
Bump golang.org/x/crypto and golang.org/x/sys

acts-1631:
Bumps golang.org/x/crypto 0.53.0 -> 0.54.0 and golang.org/x/sys
0.46.0 -> 0.47.0. These are the two golang.org/x modules dependabot
didn't open PRs for (circl #3257, miekg/dns #3261, kardianos/service
#3265 and powerman/check #3266 cover the rest).

Running go mod tidy also drops the unused github.com/miekg/dns module
and a few stale indirect deps. That same cleanup is in #3257, so
whichever merges first the other should rebase cleanly.

Vendor tree regenerated. Build, go vet and the test suite pass locally.

It would also be appreciated if a new release could be cut whenever
Go security fixes need to land, so downstream users can pick them up
without having to build from source.


jedisct1:
None of the security issues affecting the current dependencies are relevant to DNSCrypt-proxy. 

With supply chain attacks being all the rage, I can't merge such pull requests. Hope you understand.

acts-1631:
Okay, no problem. I'll still ask that you please consider keeping the deps up to date yourself then as we follow new Go releases with regular security updates. Cutting a dnscrypt-proxy release when affected dependencies are vulnerable would be much appreciated for distros. Thanks for the quick reply.

PR #3300, opened 2026-07-24
Improve comment about Tor circuit sharing

a1346054:
There is no guarantee that the Tor exit node is different for isolated streams. An attempt is made to make the path through the Tor network different for isolated streams, which means that at least one node will be different, and that is not necessarily the exit node.


jedisct1:
Thank you!

alphmeric:
Tor users could just use Tor's `DNSPort` DNS server feature

PR #3305, opened 2026-08-01
Match name suffix rules on label boundaries

munzzyy:
With both `example.com` and `ads.example.com` in a blocklist, `www.xads.example.com` is not blocked. Remove the `ads.example.com` line and it is. Adding a more specific rule turns off a broader one.

`example-blocked-names.txt` says what the broad rule is supposed to cover:

```
## *.example.com            | matches example.com and all names within that zone such as www.example.com
## example.com              | identical to the above
```

Suffix rules go into a critbit trie keyed by the reversed name, and `Eval()` looks them up with `LongestPrefix()`, which returns the longest matching *byte* prefix. Reversed, `www.xads.example.com` is `moc.elpmaxe.sdax.www`. The longest key matching that is `moc.elpmaxe.sda`, i.e. `ads.example.com`, and it lands in the middle of the `xads` label, so the boundary check rejects it. That part is right. What goes wrong is the retry: it drops a label off the query instead of moving on to the shorter rules, so `moc.elpmaxe` never gets looked at. A single retry also only ever reaches one level up, so `xads.example.com` still matches but `www.xads.example.com` and anything deeper does not.

The same matcher is behind `blocked_names`, `allowed_names` and `cloaking_rules`.

The change keeps retrying against shorter rules until one lines up on a label boundary.

Before, on 610904bba816e2547b7a95cd5f8de776c1939ad1 with the new tests applied:

```
=== RUN   TestPatternMatcherSuffixLabelBoundary/a_more_specific_rule_does_not_hide_the_zone_rule
    pattern_matcher_test.go:71: Eval("www.xads.example.com") reject = false, want true
    pattern_matcher_test.go:74: Eval("www.xads.example.com") reason = , want *.example.com
=== RUN   TestPatternMatcherSuffixLabelBoundary/several_unaligned_rules_still_fall_back_to_the_zone_rule
    pattern_matcher_test.go:71: Eval("a.b.xads.example.com") reject = false, want true
    pattern_matcher_test.go:74: Eval("a.b.xads.example.com") reason = , want *.example.com
--- FAIL: TestPatternMatcherSuffixLabelBoundary (0.00s)
=== RUN   TestPatternMatcherSuffixDifferential
    pattern_matcher_test.go:145: rules [xads.net ads.xads.com xads.com xads.com], Eval("ad.xads.xads.com") = (false, ), want (true, *.xads.com)
    pattern_matcher_test.go:145: rules [ex.net ad.net ads.ex.net], Eval("xex.xads.ex.net") = (false, ), want (true, *.ex.net)
    pattern_matcher_test.go:145: rules [ex.xads.net xads.net], Eval("xa.xex.xads.net") = (false, ), want (true, *.xads.net)
    pattern_matcher_test.go:145: rules [ads.xex.net example.net xads.com ads.xads.com], Eval("xa.xads.xads.com") = (false, ), want (true, *.xads.com)
    pattern_matcher_test.go:145: rules [xads.net ads.ex.net ex.net], Eval("xads.xads.ex.net") = (false, ), want (true, *.ex.net)
    pattern_matcher_test.go:152: 9 of 20000 generated cases disagree with the reference implementation
--- FAIL: TestPatternMatcherSuffixDifferential (0.04s)
FAIL
```

After:

```
--- PASS: TestPatternMatcherSuffixLabelBoundary (0.00s)
    --- PASS: TestPatternMatcherSuffixLabelBoundary/zone_rule_matches_a_name_in_the_zone (0.00s)
    --- PASS: TestPatternMatcherSuffixLabelBoundary/a_more_specific_rule_does_not_hide_the_zone_rule (0.00s)
    --- PASS: TestPatternMatcherSuffixLabelBoundary/the_most_specific_matching_rule_wins (0.00s)
    --- PASS: TestPatternMatcherSuffixLabelBoundary/several_unaligned_rules_still_fall_back_to_the_zone_rule (0.00s)
    --- PASS: TestPatternMatcherSuffixLabelBoundary/a_rule_is_not_matched_in_the_middle_of_a_label (0.00s)
    --- PASS: TestPatternMatcherSuffixLabelBoundary/a_rule_matches_the_zone_apex_itself (0.00s)
=== RUN   TestPatternMatcherSuffixDifferential
    pattern_matcher_test.go:154: 20000 generated cases, no disagreement with the reference implementation
--- PASS: TestPatternMatcherSuffixDifferential (0.05s)
PASS
```

The differential test builds random rule sets from labels that overlap on byte boundaries (`ads` inside `xads`) and compares `Eval()` against a naive implementation of the documented semantics.

Full suite, `gofmt` and `go vet`:

```
$ make test
cd dnscrypt-proxy && go test  -tags "" ./...
ok  	github.com/dnscrypt/dnscrypt-proxy/dnscrypt-proxy	2.791s

$ make vet
cd dnscrypt-proxy && go vet ./...

$ gofmt -l dnscrypt-proxy/
```

The loop runs more than once only when a rule matches byte-wise without landing on a label boundary. I benchmarked `Eval()` against a matcher holding 100k random rules with queries that match nothing: 29.1 µs/op before, 29.4 µs/op after.

### DIFF
--- a/dnscrypt-proxy/pattern_matcher.go
+++ b/dnscrypt-proxy/pattern_matcher.go
@@ -127,21 +127,20 @@ func (patternMatcher *PatternMatcher) Eval(qName string) (reject bool, reason st
 		return true, qName, xval
 	}
 
+	// Suffix rules are stored reversed, so the longest matching byte prefix is
+	// not necessarily aligned on a label boundary. When it isn't, retry against
+	// the shorter rules instead of giving up. Rules are never empty, so the
+	// candidate always shrinks.
 	revQname := StringReverse(qName)
-	if match, xval, found := patternMatcher.suffixes.LongestPrefix([]byte(revQname)); found {
+	for candidate := revQname; len(candidate) > 0; {
+		match, xval, found := patternMatcher.suffixes.LongestPrefix([]byte(candidate))
+		if !found {
+			break
+		}
 		if len(match) == len(revQname) || revQname[len(match)] == '.' {
 			return true, "*." + StringReverse(string(match)), xval
 		}
-		if len(match) < len(revQname) && len(revQname) > 0 {
-			if i := strings.LastIndex(revQname, "."); i > 0 {
-				pName := revQname[:i]
-				if match, xval2, found := patternMatcher.suffixes.LongestPrefix([]byte(pName)); found {
-					if len(match) == len(pName) || pName[len(match)] == '.' {
-						return true, "*." + StringReverse(string(match)), xval2
-					}
-				}
-			}
-		}
+		candidate = candidate[:len(match)-1]
 	}
 
 	if match, xval, found := patternMatcher.prefixes.LongestPrefix([]byte(qName)); found {
--- a/dnscrypt-proxy/pattern_matcher_test.go
+++ b/dnscrypt-proxy/pattern_matcher_test.go
@@ -1,0 +1,156 @@
+package main
+
+import (
+	"math/rand"
+	"strings"
+	"testing"
+)
+
+func TestPatternMatcherSuffixLabelBoundary(t *testing.T) {
+	tests := []struct {
+		name       string
+		rules      []string
+		qName      string
+		wantReject bool
+		wantReason string
+	}{
+		{
+			name:       "zone rule matches a name in the zone",
+			rules:      []string{"example.com"},
+			qName:      "www.xads.example.com",
+			wantReject: true,
+			wantReason: "*.example.com",
+		},
+		{
+			name:       "a more specific rule does not hide the zone rule",
+			rules:      []string{"example.com", "ads.example.com"},
+			qName:      "www.xads.example.com",
+			wantReject: true,
+			wantReason: "*.example.com",
+		},
+		{
+			name:       "the most specific matching rule wins",
+			rules:      []string{"example.com", "ads.example.com"},
+			qName:      "www.ads.example.com",
+			wantReject: true,
+			wantReason: "*.ads.example.com",
+		},
+		{
+			name:       "several unaligned rules still fall back to the zone rule",
+			rules:      []string{"example.com", "ads.example.com", "s.example.com"},
+			qName:      "a.b.xads.example.com",
+			wantReject: true,
+			wantReason: "*.example.com",
+		},
+		{
+			name:       "a rule is not matched in the middle of a label",
+			rules:      []string{"ads.example.com"},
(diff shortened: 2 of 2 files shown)


jedisct1:
Thank you!

The lookups were capped to 2 in order to avoid excessive CPU usage, but rather than making the number of lookups unbounded, we can simply add an extra dot. Then a key ending inside `xads` cannot be a prefix because its added dot differs from the next label byte. This matches how it's always been implemented server-side.

munzzyy:
That's better than what I sent. One lookup instead of a loop, and the boundary falls out of the key rather than being re-checked after the fact.

I ran the tests from this PR against 0de84ad05b:

```
--- PASS: TestPatternMatcherSuffixLabelBoundary (0.00s)
    --- PASS: .../zone_rule_matches_a_name_in_the_zone
    --- PASS: .../a_more_specific_rule_does_not_hide_the_zone_rule
    --- PASS: .../the_most_specific_matching_rule_wins
    --- PASS: .../several_unaligned_rules_still_fall_back_to_the_zone_rule
    --- PASS: .../a_rule_is_not_matched_in_the_middle_of_a_label
    --- PASS: .../a_rule_matches_the_zone_apex_itself
=== RUN   TestPatternMatcherSuffixDifferential
    pattern_matcher_test.go:154: 20000 generated cases, no disagreement with the reference implementation
--- PASS: TestPatternMatcherSuffixDifferential (0.04s)
```

The apex one is worth noting: `example.com` still matches itself only because the query gets the extra dot too. Appending it to the rule alone would have broken that, and your commit does both.

Want the tests on their own? They don't depend on how the matching is done. The differential one builds random rule sets out of labels that overlap on byte boundaries, like `ads` inside `xads`, and compares against a naive implementation of the documented semantics, so it would catch a regression here either way.

jedisct1:
Test added in 0acc6a57add58ac0c6405f4d8eb6d863dec0cb50

munzzyy:
Nice, that covers it end to end.

Asserting `*.example.org` in blocked-names.log is the good part — that's what catches the trailing dot leaking into the reason if the strip in `Eval()` ever drifts, which a pass/fail check on the query alone would miss.

jedisct1:
Feels weird to talk to an AI, though.

munzzyy:
Fair enough. There is a person on this end.

I use AI to get through unfamiliar code faster than I could on my own, but I read what it turns up, I run it, and I don't post anything I can't explain myself. The editcap round trip and the differential run were real.

I'll write my own from here. Sorry for the noise on a PR you had already closed.

PR #3312, opened 2026-08-11
Add `wph` and `wp<N>` strategy

ghost:


jedisct1:
It doesn't seem very useful.

PR #3313, opened 2026-08-13
security: reject HTTP redirects

rustytrees:
## Summary

- Stop the shared HTTP client from following redirects.
- Return redirect status responses as errors.
- Cover source downloads and encrypted DNS GET and POST requests.
- Test all redirect status codes followed by Go: 301, 302, 303, 307, and 308.
- Test cross-origin redirects, HTTPS downgrades, POST redirects, and direct requests.
- Verify that redirect targets are never contacted.

## Validation

```sh
go test -race ./dnscrypt-proxy
CGO_ENABLED=0 go vet ./dnscrypt-proxy
git diff --check
```

All checks passed.

## Behavior change

Endpoints that rely on HTTP redirects must be configured with their final URL.

### DIFF
--- a/dnscrypt-proxy/xtransport.go
+++ b/dnscrypt-proxy/xtransport.go
@@ -722,8 +722,9 @@ func (xTransport *XTransport) Fetch(
 		timeout = xTransport.timeout
 	}
 	client := http.Client{
-		Transport: xTransport.transport,
-		Timeout:   timeout,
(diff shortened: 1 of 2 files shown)


rustytrees:
@jedisct1 Sorry I'm new to GOLANG [this](https://jonathanmh.com/p/tracing-preventing-http-redirects-golang/) is the guide I used?

Issue #1437, opened 2020-08-04
Outgoing Network interfaces


dtouzeau:
## Is your feature request related to a problem?
When using a server with multiples Interfaces and same metrics, DNScypt-proxy chose the first available that is not the right interface for outgoing requests.

## Describe the solution you'd like
Adding a paremeter 

```
## List of local addresses and ports to send outgoing requests.
##  Can be IPv4 and/or IPv6.
## Example with both IPv4 and IPv6:
outgoing_addresses = ['127.0.0.1:53', '[::1]:53']
```

welwood08:
Looks like the second duplicate of #1362 in 24 hours. Thank you for the report but please do search for similar issues first.

jedisct1:
Maybe not a duplicate.

The way I understand this it that it's a feature request about having a way to choose the source IP address for connections to upstream servers.

welwood08:
My bad, having re-read it that makes sense. I guess it would need the same technique to implement at least.

dtouzeau:
> 
> 
> My bad, having re-read it that makes sense. I guess it would need the same technique to implement at least.

Yes you understand the feature request...

PR #3319, opened 2026-08-21
Add outgoing_interface to bind the source address of outgoing queries

nopoz:
On a firewall the OS routes by destination, not source. Multi-WAN and VPN policy routing works by matching a known source address and sending it to a chosen gateway. dnscrypt-proxy currently lets the OS pick the source for every outgoing query, so its traffic cannot be selected by such a rule and always follows the default route.

This adds an `outgoing_interface` option to bind that source. Empty by default, so nothing changes for existing installations.

Requested in #1437 (closed without implementation) and #3248.

## Configuration

```toml
## Source address for outgoing queries.
## Accepts an interface name, or a literal IP address to pick a specific
## address when an interface has several. Empty means the OS decides.
## Forwarding rules, the network probe, connections through a proxy, and
## queries to loopback destinations are all left unaffected.

# outgoing_interface = 'eth0'
```

The value is parsed as an IP first; if that fails it is treated as an interface name. No interface name parses as an IP, so there is no ambiguity.

An interface name resolves to its current addresses and is re-resolved on network change, which is what makes a DHCP lease renewal survivable. The literal form exists for selecting one address when an interface carries several, such as a CARP VIP.

I went with resolving to an address rather than `SO_BINDTODEVICE` because FreeBSD has no equivalent (`IP_SENDIF` was proposed in 2004 and never landed), and because policy routing rules match on source IP anyway, so the address is what the use case actually needs. A second Linux-only mechanism would fork the behaviour per platform for no gain.

## Scope

Bound: DNSCrypt UDP and TCP, certificate probes, bootstrap resolution, the cloaking and DNS64 name lookups, DoH and ODoH, DoH3.

Not bound, deliberately:

- `forwarding_rules`, which commonly target LAN or localhost resolvers a WAN source would make unreachable.
- The network probe, which is a pre-startup connectivity check.
- Loopback and unspecified destinations, at every site. That traffic is host-local and never traverses an interface, so binding it achieves nothing while breaking the proxy's own internal resolution, which uses `listen_addresses` as its resolver list.
- Anything going through `proxy`, where a local bind has no meaning. Setting both warns at startup.

A configured source with no address of the destination's family fails the query rather than dialling unbound, so a misconfiguration cannot silently leak to the default gateway.

## Notes

Two things that are easy to get wrong and are worth a second look during review:

- `net.Dialer.LocalAddr` is an interface field, so the concrete type has to match the network. A `*net.TCPAddr` on a `udp` network is rejected with `mismatched local address type`.
- `dns.NewTransport()` returns a struct copy whose `Dialer` field aliases a package global, so it has to be replaced before binding or every `dns.Client` in the process is affected, including the forwarding plugin's.

## Verification

Unit tests cover the resolution, family selection and fail-closed logic. No unit test can prove a packet left with the intended source, so I also built a container harness that runs the real binary against two networks and asserts the source address observed by the upstream. It covers the option unset, the interface-name and literal forms, the DNSCrypt path through the connection pool, an invalid interface name, the fail-closed case, and the loopback exclusion.

Each of those checks was confirmed to fail when the corresponding logic is reverted, rather than only confirmed to pass. I can share the harness separately if desired; I left it out here to keep the diff to the feature.

### On pfSense

Also verified on pfSense 2.8.1-RELEASE (FreeBSD 15.0-CURRENT, amd64), since the FreeBSD reasoning above is the whole basis for the design. The proxy and the test binary were cross-compiled with `CGO_ENABLED=0 GOOS=freebsd GOARCH=amd64` and run on the firewall as a second instance on a spare loopback port, so nothing serving real DNS was touched.

- The unit suite runs natively: 27 tests, all passing, none skipped.
- `outgoing_interface = 'ix1.5'` moved the observed source from `10.0.0.1` to `10.0.2.1`, so interface-name resolution works, VLAN names included.
- The literal form selected a secondary address on an interface carrying two, which is the CARP VIP case.
- `outgoing_interface = 'zzz0'` aborted at startup with `Unknown outgoing interface [zzz0]: route ip+net: no such network interface`.
- The bound source reached the wire unrewritten on the WAN interface. Whether it survives at all depends on the outbound NAT rules in play, which is the caveat below rather than a property of the option.

This option sets the source address and nothing else. It does not select a route or a gateway. On a firewall it makes DNS traffic distinguishable so that policy routing can act on it, which is the point, but the routing itself remains the operator's job. Outbound NAT can also rewrite a bound source before it leaves the box, so a matching NAT exemption may be needed for the binding to be observable.

### DIFF
--- a/dnscrypt-proxy/config.go
+++ b/dnscrypt-proxy/config.go
@@ -97,6 +97,7 @@ type Config struct {
 	TLSKeyLogFile            string                      `toml:"tls_key_log_file"`
 	NetprobeAddress          string                      `toml:"netprobe_address"`
 	NetprobeTimeout          int                         `toml:"netprobe_timeout"`
+	OutgoingInterface        string                      `toml:"outgoing_interface"`
 	OfflineMode              bool                        `toml:"offline_mode"`
 	HTTPProxyURL             string                      `toml:"http_proxy"`
 	RefusedCodeInResponses   bool                        `toml:"refused_code_in_responses"`
--- a/dnscrypt-proxy/config_loader.go
+++ b/dnscrypt-proxy/config_loader.go
@@ -96,6 +96,18 @@ func configureXTransport(proxy *Proxy, config *Config) error {
 	proxy.xTransport.useIPv6 = config.SourceIPv6
 	proxy.xTransport.keepAlive = time.Duration(config.KeepAlive) * time.Second
 
+	outgoing, err := newOutgoingSource(config.OutgoingInterface)
+	if err != nil {
+		return err
+	}
+	if outgoing != nil && (len(config.Proxy) > 0 || len(config.HTTPProxyURL) > 0) {
+		dlog.Warn("outgoing_interface has no effect on connections made through a proxy")
+	}
+	proxy.xTransport.outgoing = outgoing
+	if outgoing != nil {
+		outgoing.onRefresh = proxy.udpConnPool.Flush
+	}
+
 	// Configure HTTP proxy URL if specified
 	if len(config.HTTPProxyURL) > 0 {
 		httpProxyURL, err := url.Parse(config.HTTPProxyURL)
--- a/dnscrypt-proxy/dnsutils.go
+++ b/dnscrypt-proxy/dnsutils.go
@@ -512,7 +512,7 @@ func _dnsExchange(
 			proxy.prepareForRelay(udpAddr.IP, udpAddr.Port, &binQuery)
 			upstreamAddr = relay.RelayUDPAddr
 		}
-		pc, err := net.DialTimeout("udp", upstreamAddr.String(), proxy.timeout)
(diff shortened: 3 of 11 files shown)


jedisct1:
This is a large change, and it touches critical parts of the code (if networking stops working, nothing works).

So I'd rather implement the feature manually and incrementally, so it can be understood and maintained by humans.

nopoz:
@jedisct1 I get wanting to be careful, but why close the PR right away with no discussion? I was very thorough in my testing and validation. Unconfigured, the current default behavior stays the same so the blast radius should be minimal.

jedisct1:
This is very useful!

It gives ideas and concrete things to consider for that feature.

But it is too big of a change to merge as-is.

jedisct1:
The `outboundip` is an attempt to add that feature.

But the changes are intrusive and will be painful to maintain.

I'm sorry, I don't think that feature can be added.

